PR #35, opened 2014-03-11
patch nodejs to use python2 instead of system "python"

mfussenegger:
system "python" may refer to either python2 or python3 and nodejs will fail to
compile if it is python3

see http://legacy.python.org/dev/peps/pep-0394/


celaus:
stole your changes to my branch :)

PR #95, opened 2015-01-12
fixed cluster view render issue in safari

chaudum:


seut:
please check if watcher is running or/and using resources when leaving the page.
otherwise ack

chaudum:
watchers are automatically destroyed when switching "pages"

seut:
ack

PR #125, opened 2015-06-11
crate process graph update

mikethebeer:
… process


chaudum:
pls add changelog entry in CHANGES.txt

chaudum:
ack

PR #133, opened 2015-06-12
feature: plugins

chaudum:
load plugins.json before bootstrapping angular app to be able to inject additional modules on start


mikethebeer:
ack

PR #135, opened 2015-06-19
fixes

chaudum:
- fix: automatically refresh load history chart on overview page
 - fix: hide navigation item title when screen width is less than 1200px


mikethebeer:
ack

PR #137, opened 2015-06-22
approx number of underrepl/missing records was calculated incorrectly

chaudum:
which lead to different results when viewing on admin ui on different nodes

### DIFF
--- a/app/scripts/services/tableinfo.js
+++ b/app/scripts/services/tableinfo.js
@@ -2,90 +2,115 @@
 
 angular.module('tableinfo', ['sql'])
   .factory('TableInfo', function() {
-    return function TableInfo(shards, numConfigured, partitionedBy) {
-      this.partitionedBy = partitionedBy || [];
-      this.partitioned = this.partitionedBy.length > 0;
-      this.shards = shards;
-      this.shards_configured = numConfigured || 0;
-      this.primaryShards = function primaryShards() {
-        return this.shards.filter(function (shard, idx) {
+    return function TableInfo(shards, configured, partitionedBy) {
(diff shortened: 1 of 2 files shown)


app/scripts/services/tableinfo.js
```diff
@@ -2,90 +2,115 @@
...
+      var numUnderreplicatedRecords = (function(){
+        var avgDocsPerPrimaryShard = primaryShards.reduce(function(memo, shard, idx, arr){
+          return memo + shard.avg_docs / arr.length;
+        }, 0);
+        return avgDocsPerPrimaryShard * numUnderreplicatedShards;
+      }());
+
+      var numUnavailableRecords = (function() {
+        var avgDocsPerStartedShard = startedShards.reduce(function(memo, shard, idx, arr){
+          return memo + shard.avg_docs / arr.length;
```
chaudum: this is the actual fix

app/scripts/services/tableinfo.js
```diff
@@ -2,90 +2,115 @@
...
+        return shards.filter(function(shard, idx) {
+          return shard.state == 'UNASSIGNED';
+        }).reduce(function(memo, shard, idx) {
+          return memo + shard.count;
+        }, 0);
+      }());
+
+      var numUnderreplicatedRecords = (function(){
+        var avgDocsPerPrimaryShard = primaryShards.reduce(function(memo, shard, idx, arr){
+          return memo + shard.avg_docs / arr.length;
```
chaudum: this is the actual fix

mikethebeer:
ack

mikethebeer:
ack

PR #141, opened 2015-08-13
show relative Crate process CPU usage

chaudum:
and display number of cores per node

Note: this feature will require Crate 0.52.x


kovrus:
ack

PR #151, opened 2015-10-19
fix cluster not reachable

mikethebeer:
* hide console

### DIFF
--- a/app/scripts/controllers/console.js
+++ b/app/scripts/controllers/console.js
@@ -106,7 +106,6 @@ angular.module('console', ['sql'])
             $scope.rows = sqlQuery.rows;
             $scope.status = sqlQuery.status();
             $scope.statement = stmt + ";";
-
             inputScope.updateInput($scope.statement);
           })
           .error(function(sqlQuery) {
--- a/app/scripts/services/stats.js
+++ b/app/scripts/services/stats.js
@@ -41,6 +41,8 @@ angular.module('stats', ['sql', 'health', 'tableinfo'])
           }
           setReachability(false);
         }
+      }).error(function(data, status) {
+        setReachability(false);
       });
     };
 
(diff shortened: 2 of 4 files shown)


app/scripts/controllers/console.js
```diff
@@ -220,4 +220,10 @@ angular.module('console', ['sql'])
     };
   })
   .controller('ConsoleController', function ($scope, $http, $location, SQLQuery, $log, $timeout) {
+    var clusterName = SQLQuery.execute('select name from sys.cluster');
+    clusterName.success(function(sqlQuery) {
+        $scope.online = true;
+    }).error(function(data, status) {
+      $scope.online = false;
```
chaudum: code style: same indentation

app/scripts/services/stats.js
```diff
@@ -135,10 +137,11 @@ angular.module('stats', ['sql', 'health', 'tableinfo'])
     };
 
     var refreshHealth = function() {
-      if (!data.online) return;
       // We want to get the tables as soon as they become available so we use the promise object.
       TableList.promise.then(null, null, function(res){
         if (res.success) {
```
chaudum:
```js
if (res.success || !data.online) { ... }
```

app/views/console.html
```diff
@@ -16,7 +16,7 @@
           </div>
 
 
-          <form role="form">
+          <form role="form" ng-show="online">
```
chaudum: I would not hide the console, but catch the case if someone wants to submit a query.

app/views/console.html
```diff
@@ -43,7 +43,7 @@
             </div>
 
           </form>
-        </div>
```
chaudum: ?

chaudum:
ack

PR #154, opened 2015-10-28
Make overview page show correct % of replicated data

kovrus:


app/scripts/services/tableinfo.js
```diff
@@ -80,6 +80,17 @@ angular.module('tableinfo', ['sql'])
         return Math.ceil(avgDocsPerPrimaryShard * numUnderreplicatedShards);
       }());
 
+       var numRecordsWithReplicas = (function() {
+        var avgDocsPerPrimaryShard = primaryShards.reduce(function(memo, shard, idx, arr) {
```
chaudum: move that out of the function scope, because it can be used for the `numUnderreplicatedRecords` method too

chaudum:
ack

PR #155, opened 2015-10-28
Return correct number of primary shards

kovrus:


chaudum:
ack

chaudum:
you could also change it here:
https://github.com/crate/crate-admin/pull/155/files#diff-681ec8c017071c97970e1afb9d4b0f7aR36
filter the `primaryShards` instead of using `shards` and filter for `primary === true`

chaudum:
ack

PR #159, opened 2015-11-03
UDC: identify anonymous user with cluster id.

chaudum:
Both user id and cluster id are anonymous traits.
You can disable UDC via the `udc.enabled` setting on the Crate server.


celaus:
ack. Squash and merge :+1:

PR #160, opened 2015-11-06
fix: charts are not correctly displayed in ms edge

mikethebeer:


CHANGES.txt
```diff
@@ -5,6 +5,8 @@ Changes for Crate Admin Interface
 Unreleased
 ==========
 
+ - fix: charts are not correctly displayed when using microsoft edge browser.
```
chaudum: when using -> in

chaudum:
otherwise ack

PR #161, opened 2015-11-09
version check was always fetched with `udc.enabled=false`

chaudum:
even when the `udc.enabled` setting was enabled on the Crate server


mikethebeer:
ack

PR #165, opened 2015-11-13
improved table list filter

chaudum:


celaus:
Please also remove the now unused model "tableName" in tablelist controller.

celaus:
After that: Ack ;)

PR #166, opened 2015-11-16
prepare release 0.16.0

mikethebeer:


CHANGES.txt
```diff
@@ -5,6 +5,9 @@ Changes for Crate Admin Interface
 Unreleased
 ==========
 
+2015/11/16 0.16.0
+=================
+
  - Improved table list filter: maintain default sort order (health, name)
```
chaudum: this is a fix to the previous changelog entry. must be taken out!

chaudum:
also the fix:
```
  - fix: Crate version check was always fetched with `udc.enabled=false`
    even when the `udc.enabled` setting was enabled on the Crate server.
```
is part of the new UDC feature. so must be removed too

kovrus:
ack

PR #168, opened 2015-11-26
updated calculation of underreplicated shards/records to work with newly

chaudum:
exposed shard states and relocating/recovering shards in `sys.shards` table


celaus:
ack!

PR #171, opened 2015-11-27
fix: column headers where not shown in result table on console page

chaudum:
when two or more columns with the same name were selected

screen shot 2015-11-27 at 13 09 23


kovrus:
ack

PR #180, opened 2016-01-21
include unassigned shards in calculation of recovery percentage

chaudum:


app/scripts/services/shardinfo.js
```diff
@@ -84,7 +84,7 @@ angular.module('shardinfo', [])
         SQLQuery.execute(recoveryStmt)
           .success(function (recoveryQuery) {
             var result = queryResultToObjects(recoveryQuery,
-                ['table_name', 'schema_name', 'recovery_stage', 'recovery_percent']);
+                ['table_name', 'schema_name', 'recovery_stage', 'recovery_percent', 'count']);
```
seut: afaik recovery_percent is not available anymore, you've replaced it with avg(recovery['size']['percent'])

seut:
ack. maybe changes can be improved (though I don't have an suggestion ;-)

PR #181, opened 2016-01-21
fix: change in calculation of recovery percentage

chaudum:
broke calculations for partitioned tables causing table partitions not showing up


mikethebeer:
ack

PR #183, opened 2016-01-26
do not count relocating shards as "underreplicated" shards

chaudum:
relocating shards where previously displayed as "underreplicated" shards
causing a yellow cluster state while relocating was in progress


seut:
ack

PR #189, opened 2016-02-02
updated Bootstrap to latest 3.x version

chaudum:
this required grunt-contrib-less instead of deprecated grunt-recess


celaus:
we were not using grunt-contrib-less?! ack

PR #193, opened 2016-04-01
use id instead of name to address nodes in

bogensberger:
cluster overview to avoid rendering errors
if there are more than one nodes with the same name

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,6 +5,10 @@ Changes for Crate Admin Interface
 Unreleased
 ==========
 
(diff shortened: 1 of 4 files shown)


CHANGES.txt
```diff
@@ -5,6 +5,9 @@ Changes for Crate Admin Interface
 Unreleased
 ==========
 
+ - Fixed an issue which caused rendering errors in the node list if the cluster
+   contains multiple nodes with the same name
```
chaudum: maybe add a note, that URLs will change due to this fix

chaudum:
ack

PR #197, opened 2016-05-19
updated analytics tokens of URLs in dropdown menu

chaudum:
see also https://github.com/crate/crate-web/pull/396


mikethebeer:
ack

PR #200, opened 2016-06-03
resolve path to ``/_sql`` endpoint correctly

chaudum:
in cases when Crate is served at a different location than ``/``, e.g. when proxied to a subdirectory such as ``/crate/`` then also the ``_sql`` endpoint is located at ``/crate/_sql``.
This change removes the hard coded path and resolves the correct location of ``_sql`` based on the relative location to ``/_plugin/crate-admin/``.


mikethebeer:
ack

PR #203, opened 2016-06-13
add support of i18n for es2015 branch

Kaijun:
Hi, 

i saw you guys are now working on the es2015 branch, 
So I just open a new branch based on es2015 for adding i18n support by using [angular-translate](https://github.com/angular-translate/angular-translate)

- [x] Extract almost all encoded strings from templates as well as some messages in controllers.
- [x] standalone json file in `i18n` folder for each language
- [x] a sweet button at the top right conner of the status bar for switching different languages
- [ ] full translation (now only part of chinese and german was done)
- [ ] probably a better naming strategy for each string 

the main configuration of this functionality is in `app.js`
```
    // Configuration of i18n Internationalization
    app.config(['$translateProvider', 
      function($translateProvider) {

        // configures staticFilesLoader
        $translateProvider.useStaticFilesLoader({
          prefix: '/i18n/',
          suffix: '.json'
        })
        .registerAvailableLanguageKeys(['en', 'de', 'zh'], {
          'en_*': 'en',
          'de_*': 'de',
          'zh_*': 'zh'
        })
        // Check local language automatically
        .determinePreferredLanguage()
        // Use English as default language if no matched language is found.
        .fallbackLanguage('en');

      }]);
```

after the configuration, all hard-coded strings in views are replaced in `angular-translate` way, like:
```
{{ 'OVERVIEW.INFO' | translate }}
```
`OVERVIEW.INFO` will be replaced later with the string value `Overview/Übersicht/概览` defined in json files in `i18n` folder.

😄


chaudum:
We'll have a look as soon as possible!

app/scripts/controllers/common.js
```diff
@@ -87,6 +87,24 @@ var commons = angular.module('common', ['stats', 'udc'])
...
+
+      var queryElements = this.navBarElements.filter(function(item) {
+        return item.urlPattern==urlPattern;
+      });
+
+      if(queryElements.length===0){
+        return;
+      }
+
+      var element = queryElements[0];
```
chaudum: no need to allocate new variable

```js
queryElements[0].text = text;
```

app/views/node.html
```diff
@@ -88,35 +91,35 @@ <h3 class="panel-headline">Disk Usage</h3>
...
               <span class="load-value">{{ node.load['1']|number:2 }}</span>
               <span class="load-value">{{ node.load['5']|number:2 }}</span>
               <span class="load-value">{{ node.load['15']|number:2 }}</span>
             </p>
             <p class="announcement-heading" ng-hide="node.load['1'] >= 0">-</p>
             <p class="announcement-text">
-              <span class="load-label">1min</span>
-              <span class="load-label">5min</span>
-              <span class="load-label">15min</span>
+              <span class="load-label">1{{ 'NODE.MIN' | translate }}</span>
```
chaudum: I would name the key `NODE.MINUTES` because `min` could either mean `minimum` or `minutes`. I our case `minutes`.

chaudum: Or even just `COMMON.MINUTES` since it's not node specific.

chaudum:
@Kaijun the PR looks really good! thanks for the work.
we could even provide the German translation if you want.
and could you please open the PR against `master` since `es2015` is WIP and won't get merged soonish.

chaudum:
I discovered that when you open the language dropdown menu and move the cursor over the text it changes from pointer to selector.
However, I think it should behave the same way as the "help" dropdown behaves (with hand cursor).

chaudum:
the language selection is not persisted. whenever you refresh the admin ui it will detect the browser language again. I think there should be an item "Autodetect Language" in the dropdown that clears the previously persisted language value

chaudum:
I've done the German translation. I would push it to the new PR against master.

chaudum:
hi @Kaijun 
any news on this? do you want us to finish the implementation?
best, christian

Kaijun:
sry was busy yesterday. I'll send a new pr late today. Thanks!

Kaijun:
@chaudum new [PR](https://github.com/crate/crate-admin/pull/205)  was opened. 🎉

chaudum:
merged into master with #205

PR #204, opened 2016-06-17
trim trailing slash from base_uri parameter

chaudum:
in order to also allow urls with trailing slash, such as
`http://localhost:4200/` as valid `base_uri` parameter for local
development


bogensberger:
ack

PR #221, opened 2016-09-22
use storedURI properly to display log message

m-kharbat:
@chaudum please review

### DIFF
--- a/app/scripts/services/stats.js
+++ b/app/scripts/services/stats.js
@@ -33,9 +33,6 @@ angular.module('stats', ['sql', 'health', 'tableinfo', 'nodeinfo'])
           setReachability(true);
         } else {
           data.version = null;
-          if (!storedURI) {
-            $log.warn('If you develop and run Crate Admin UI locally you need to set the base_uri. See README.rst for further information.');
-          }
           setReachability(false);
         }
       }).error(function(data, status) {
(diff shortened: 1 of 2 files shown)


app/scripts/services/stats.js
```diff
@@ -33,6 +33,7 @@ angular.module('stats', ['sql', 'health', 'tableinfo', 'nodeinfo'])
           setReachability(true);
         } else {
           data.version = null;
+          var storedURI = localStorage.getItem("crate.base_uri");
```
chaudum: you need to use `baseURI` here
see `sql.js`

m-kharbat: baseURI will return the current host if there is nothing stored in the localStorage. It doesn't allow you to check if there is something stored there or not .. unless you want to move this log message to the BaseURI factory

```
  .factory('baseURI', function($location, $log) {
    return function baseURI(path) {
      var basePath = localStorage.getItem("crate.base_uri");
      if (!basePath) {
        $log.warn('...'); // <----
        var pluginPath = '/_plugin/crate-admin/';
        basePath = $location.protocol() + "://" +
          $location.host() + ":" +
          $location.port() +
          window.location.pathname.replace(pluginPath, '');
      }
      // remove trailing slash from base path and append path
      return basePath.replace(/\/+$/, '') + path;
    };
  })
```

chaudum: yes, we should move the warning into the `baseURI` function
this should be the single place to obtain the URI from the storage

PR #222, opened 2016-09-22
some translation fixes that bugged me 😬

celaus:
Also, y'all need to work on your german skills :P


app/views/tables.html
```diff
@@ -10,7 +10,7 @@
         <div class="col-sm-6 col-xs-12">
           <div class="panel panel-info">
             <div class="panel-heading">
-              <p class="announcement-heading"><b>{{ table_label }}</b> <span ng-show="isParted" class="small">(partitioned)<span></p>
+              <p class="announcement-heading"><b>{{ (nr_of_tables == 0 && ('TABLE.TITLE' | translate) + ' (' + nr_of_tables + ')') || table_label }}</b> <span ng-show="isParted" class="small">(partitioned)<span></p>
```
m-kharbat: @celaus i'm not sure I understand what this line does

celaus: a && b || c or, in Python: myvar  = a and b or c where a, b, and c are expressions. a is the boolean condition that should apply, and if it does, b also has to be true (in the JS/Python sense) to be returned to the caller (this is a language feature. In Java it would only return 'true'). If a evaluates to false, short-circuit evaluation means that if the first part (a) is already false, just skip ahead and return whatever is the second part of the OR, since this will be the outcome. 

ehm. makes sense? I don't like the solution either, maybe there's something better.

app/i18n/de.json
```diff
@@ -92,28 +92,28 @@
         "TYPE": "Typ"
     },
     "NODE": {
-        "NOT_REACHABLE_MSG": "Cluster ist nicht erreichbar",
+        "NOT_REACHABLE_MSG": "Kein Cluster erreichbar",
```
chaudum: I question this translation:
Kein Cluster erreichbar = No cluster available
Cluster nicht erreichbar = Cluster not available

app/views/tables.html
chaudum: Isn't there pluralisation for Angular?

app/views/tables.html
```diff
@@ -10,7 +10,14 @@
         <div class="col-sm-6 col-xs-12">
           <div class="panel panel-info">
             <div class="panel-heading">
-              <p class="announcement-heading"><b>{{ table_label }}</b> <span ng-show="isParted" class="small">(partitioned)<span></p>
+              <p class="announcement-heading"><b>
+                <ng-pluralize ng-init="translationKey='TABLE.TITLE'" count="nr_of_tables"
```
chaudum: crazy shit

celaus: This is how we roll

PR #234, opened 2016-10-04
fix: history doesn't work properly

m-kharbat:
two things were not working properly:

1. if the index of the recentQueries is too high it returned "undefined" and resulted in this error:
screen shot 2016-10-04 at 13 37 23

2. the recentQueries array was always inconsistent and didn't store the last query typed (in the sa,e session)

@chaudum please review

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,6 +5,8 @@ Changes for Crate Admin Interface
 Unreleased
 ==========
 
+ - Fixed an issue that caused the query history in the console view to be inconsistent.
+
  - Added Spanish translation.
 
 2016/09/23 0.20.0
--- a/app/scripts/controllers/console.js
+++ b/app/scripts/controllers/console.js
@@ -5,15 +5,16 @@ angular.module('console', ['sql'])
     return {
       restrict: 'A',
       controller: ['$scope', '$translate', function($scope, $translate){
+        var self = this;
 
         var inputScope = null;
         var rows = [];
         var resultHeaders = [];
         var renderTable = false;
-        var recentQueries = [];
         var statement = "";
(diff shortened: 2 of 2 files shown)


app/scripts/controllers/console.js
```diff
@@ -207,21 +206,24 @@ angular.module('console', ['sql'])
...
+          var recentQueriesIndex = 0;
           if (!event.shiftKey && event.keyCode === 38) {
             // UP KEY
             if ((cursor.ch === 0 && cursor.line === 0) || (cursor.line === 0 && selection === statement)) {
               if ($scope.recentCursor === 0) {
                 typedStatement = statement;
               }
               $scope.recentCursor--;
-              statement = $console.recentQueries[queryCount + $scope.recentCursor];
+              recentQueriesIndex = (queryCount + $scope.recentCursor) > 0 ? queryCount + $scope.recentCursor : 0;
```
chaudum: we do the same calculation twice. could you put that into a function call?

chaudum: I mean line `226`

CHANGES.txt
```diff
@@ -5,6 +5,8 @@ Changes for Crate Admin Interface
 Unreleased
 ==========
 
+ - Fix: console sql query history not working properly.
```
chaudum: re-phase it a bit nicer please

app/scripts/controllers/console.js
```diff
@@ -199,6 +198,13 @@ angular.module('console', ['sql'])
           $console.updateStatement(statement);
         });
 
+
+        // function to get the recentQueries index
+        function getRecentQueriesIndex(queryCount,recentCursor ){
```
chaudum: code style: whitespaces

CHANGES.txt
chaudum: 🌈

PR #237, opened 2016-10-09
use column table_schema in order to be compliant with the SQL standard.

jodok:


CHANGES.txt
```diff
@@ -5,6 +5,9 @@ Changes for Crate Admin Interface
 Unreleased
 ==========
 
+ - Use column name ``table_schema`` of ``information_schema.tables`` and
```
chaudum: this is not a user facing change so you can omit the changelog entry

chaudum:
https://github.com/crate/crate/pull/4218

jodok:
@chaudum `table_schema` is now available in crate master. let's try to include this version of the admin ui with the next release.

PR #256, opened 2016-11-02
added tests for navigationService

m-kharbat:


app/scripts/controllers/common.js
```diff
@@ -78,44 +84,55 @@ var commons = angular.module('common', ['stats', 'udc'])
...
+    var navigationService = {
+      addNavBarElement: addNavBarElement,
+      updateNavBarElement: updateNavBarElement,
+      navBarElements: navBarElements
+    };
+
+    return navigationService;
 
-    this.addNavBarElement = function(iconClass, text, urlPattern, index) {
+    ////
```
kovrus: should be removed?

m-kharbat: it's not necessary but i'm trying to follow angular guidelines for more clarity https://github.com/johnpapa/angular-styleguide/blob/master/a1/README.md#factories

PR #258, opened 2016-11-02
fix: crate tweets table query

m-kharbat:


CHANGES.txt
```diff
@@ -5,6 +5,8 @@ Changes for Crate Admin Interface
 Unreleased
 ==========
 
+ - Fix : Corrected the create tweets tabke sql query in the tutorial.
```
kovrus: tabke -> table. Maybe say smth like fixed incorrect sql query for creating the tweets table in the tutorial

PR #265, opened 2016-11-08
implement overview

m-kharbat:
screen shot 2016-11-08 at 10 46 04

screen shot 2016-11-08 at 11 01 12


chaudum:
I dislike the mixed format class names with dash and underscore, like `.cr-panel-block__checks` :(
also, please remove commented out code

PR #269, opened 2016-11-09
Implement nodelist

Autophagy:
nodelist

listlist

### DIFF
--- a/app/scripts/filter/text.js
+++ b/app/scripts/filter/text.js
@@ -38,3 +38,21 @@ angular.module('common')
       }
     };
   });
+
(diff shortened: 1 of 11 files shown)


app/scripts/filter/text.js
```diff
@@ -37,4 +37,22 @@ angular.module('common')
           return  $filter('translate')("OVERVIEW.INF'");
       }
     };
-  });
\ No newline at end of file
+  });
+
+angular.module('common')
```
m-kharbat: please also add the corresponding test for this filter

PR #273, opened 2016-11-10
Implement node

Autophagy:
nodes

### DIFF
--- a/app/styles/common.less
+++ b/app/styles/common.less
@@ -13,8 +13,20 @@ h1 {
 
 }
 
+h2, h3 {
+  font-weight: bold;
+}
+
+h3 {
+  font-weight: @font-size-h3;
+}
+
 .small {
   font-size: @font-size-small;
+}
+
+a {
+  cursor: pointer;
 }
 
 // Structure
@@ -30,7 +42,7 @@ h1 {
   position: absolute;
   top: @status-bar-height;
   left: @side-nav-width;
-  width:  calc(~"100vw -" @side-nav-width);
+  width:  calc(~"100vw -" @side-nav-width + 15px);
   height: calc(~"100vh -" @status-bar-height);
 }
 
@@ -77,6 +89,13 @@ h1 {
   padding-right: 3px;
   font-size: @font-size-medium;
   width: 200px;
+}
+
+.cr-panel-block__item__content--full-width{
+  padding-left: 3px;
+  padding-right: 3px;
+  font-size: @font-size-medium;
+  width: 100%;
 }
 
 .cr-panel--default{
@@ -188,7 +207,7 @@ h1 {
 @media only screen and (max-width:  @max-width-s){
   #page-wrapper {
     left: @side-nav-width-xs;
-    width:  calc(~"100vw -" @side-nav-width-xs);
+    width:  calc(~"100vw -" @side-nav-width-xs + 15px);
   	height: calc(~"100vh -" @status-bar-height);
   }
 }
--- a/app/styles/node.less
+++ b/app/styles/node.less
@@ -1,0 +1,108 @@
+.node-detail {
+  display: flex;
+  justify-content: space-between;
+  flex-wrap: wrap;
+}
+
+.node-detail__row {
+  display: flex;
+  justify-content: space-between;
+  width: 50%;
+}
+
+.node-detail__cell {
+  width: 50%;
+  padding: 0px 15px;
+}
+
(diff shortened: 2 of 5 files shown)


app/styles/common.less
```diff
@@ -30,7 +42,7 @@ h1 {
   position: absolute;
   top: @status-bar-height;
   left: @side-nav-width;
-  width:  calc(~"100vw -" @side-nav-width);
+  width:  calc(~"100vw -" @side-nav-width + 15px);
```
m-kharbat: this will impact all the pages using this class, did you check the other views?

app/styles/common.less
```diff
@@ -188,7 +207,7 @@ h1 {
 @media only screen and (max-width:  @max-width-s){
   #page-wrapper {
     left: @side-nav-width-xs;
-    width:  calc(~"100vw -" @side-nav-width-xs);
+    width:  calc(~"100vw -" @side-nav-width-xs + 15px);
```
m-kharbat: same comment

app/styles/node.less
```diff
@@ -0,0 +1,108 @@
...
+}
+
+.node-detail__row {
+  display: flex;
+  justify-content: space-between;
+  width: 50%;
+}
+
+.node-detail__cell {
+  width: 50%;
```
m-kharbat: what does it look like in small screen ?
if you want the elements of your flex container to have the same width, you can set flex-grow = 1

app/views/node.html
```diff
@@ -3,203 +3,285 @@
...
-          <div class="panel">
-            <p class="announcement-heading inline-block" ng-show="shardInfo.postrecovery> -1">{{ shardInfo.postrecovery }}</p>
-            <p class="announcement-heading inline-block" ng-hide="shardInfo.postrecovery> -1">-</p>
-            <p class="announcement-text">{{ 'NODE.POST_RECOVERY' | translate }}</p>
-          </div>
+        <div class="node-detail__subsection__cell--small">
+          <p class="cr-panel-block__item__header ng-binding">
+            15{{ 'NODE.MINUTES' | translate }}
+          </p>
+          <p class="cr-panel-block__item__content--full-width cr-panel--default" ng-show="node.load['1'] >= 0">
```
m-kharbat: why do you have ng-show="node.load['1'] instead of ng-show="node.load['15']

app/views/node.html
```diff
@@ -3,203 +3,285 @@
...
+            {{ node.load['1']|number:2 }}
+          </p>
+        </div>
+
+        <div class="node-detail__subsection__cell--small">
+          <p class="cr-panel-block__item__header ng-binding">
+            5{{ 'NODE.MINUTES' | translate }}
+          </p>
+          <p class="cr-panel-block__item__content--full-width cr-panel--default" ng-show="node.load['1'] >= 0">
```
m-kharbat: why do you have ng-show="node.load['1'] instead of ng-show="node.load['5'] ?

app/views/node.html
```diff
@@ -3,203 +3,285 @@
...
+          </p>
         </div>
       </div>
+    </div>
+
+    <div class="node-detail__subsection--medium">
+      <h2 class="">{{ 'NODE.SHARDS' | translate }}</h2>
+      <div class="node-detail__subsection__detail cr-panel-block">
+        <div class="node-detail__subsection__cell">
+          <p class="cr-panel-block__item__header ng-binding">
```
m-kharbat: what does ng-binding class do ?

app/styles/common.less
Autophagy: All views seem to look fine. The problem this was addressing is that when the page has a scroll bar the page is too big for the window, and results in a really ugly horizontal scrolling situation. Was unsure how to get around it other than by doing this.

app/styles/node.less
Autophagy: Small screen:

nodes

Very small:

v-small-screen

app/views/node.html
Autophagy: Only because in the original in master, the logic that determined whether any of them were displayed at all was this. See https://github.com/crate/crate-admin/blob/master/app/views/node.html#L100

app/views/node.html
Autophagy: as above ^^

app/views/node.html
Autophagy: Nothing, and im not even sure why I put this in here yesterday. I'll remove it...

PR #279, opened 2016-11-11
Implemented tutorial

Autophagy:
import

import-2

### DIFF
--- a/app/styles/common.less
+++ b/app/styles/common.less
@@ -46,7 +46,20 @@ a {
   cursor: pointer;
 }
 
-// 
+li {
+  margin: 10px 0px;
+}
+
+code {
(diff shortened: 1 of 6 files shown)


app/styles/common.less
```diff
@@ -46,7 +46,20 @@ a {
...
 
-// 
+li {
+  margin: 10px 0px;
+}
+
+code {
+  padding: 2px 4px;
+  font-size: 90%;
+  color: #c7254e;
```
m-kharbat: should this also be a variable ?

Autophagy: It's currently the same across both themes but thats a good idea - i'll include it. Can't hurt.

PR #280, opened 2016-11-11
updated overview graph to new design

chaudum:
screen shot 2016-11-11 at 14 44 08
screen shot 2016-11-11 at 14 44 14


app/styles/overview.less
```diff
@@ -1,3 +1,7 @@
+@radiobutton-bg-lightblue:  #55d4f5;
```
m-kharbat: we define the variable in "light-theme-variables.less" and "dark-theme-variables.less"

PR #293, opened 2016-11-15
Admin ui layout

m-kharbat:


app/styles/charts.less
```diff
@@ -0,0 +1,7 @@
+svg text {
+  font: 16px BlenderPro;
```
chaudum: I think the design does not use the Blender font any more.
So this is the last place where the font is used => remove font files.

app/styles/dark-theme-variables.less
```diff
@@ -0,0 +1,256 @@
+//
+// Variables
+// --------------------------------------------------
+
+//
+// Body
+// --------------------------------------------------
+
+@body-bg:               #2e2e2e;
```
chaudum: just a minor thing: align values

app/styles/main.less
```diff
@@ -3,6 +3,18 @@
 @import "../bower_components/ladda-bootstrap/dist/ladda.css";
 @import "path";
 @import "fonts";
-@import "variables";
-@import "sb-admin";
-@import "theme";
+
+@import "common";
+@import "light-theme-variables";
+// @import "dark-theme-variables";
```
chaudum: make a comment about the different themes

app/styles/main.less
```diff
@@ -3,6 +3,18 @@
 @import "../bower_components/ladda-bootstrap/dist/ladda.css";
```
chaudum: ladda is not used any more, right?

chaudum: no, it actually is....

PR #306, opened 2016-11-21
removed blender font because it is not used any more

chaudum:


m-kharbat:
for this pr can you check with @cleishm i'm not yet sure if we will get rid of this font or if we will switch to blender Pro

PR #308, opened 2016-11-22
Improved code quality by using JSHint

chaudum:
run `bin/grunt jshint`
jshint is also run before travis tests

@m-kharbat and @Autophagy please

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,8 @@ before_script:
   - "sh -e /etc/init.d/xvfb start"
   - "npm install -g bower karma grunt-cli"
   - "bower install"
-script: grunt test
+script:
+  - grunt test
 branches:
   except:
     - /.*\/.*/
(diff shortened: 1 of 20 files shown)


.travis.yml
```diff
@@ -6,7 +6,9 @@ before_script:
   - "sh -e /etc/init.d/xvfb start"
   - "npm install -g bower karma grunt-cli"
   - "bower install"
-script: grunt test
+script:
+  - grunt jshint
+  - grunt test
```
m-kharbat: instead of adding it here, it can also be added here to recognise errors locally 

```
  grunt.registerTask('server', [
    'concurrent:server',
    'copy:i18nTmp',
    'connect:dev',
    'jshint',
    'watch'
  ]);

  grunt.registerTask('test', [
    'jshint',
    'karma:all_tests'
  ]);
```

chaudum: it's a good idea to also put it also in the server tasks. however I think travis should also run it.

PR #320, opened 2016-11-28
add tests for health.js

m-kharbat:
@chaudum please review

### DIFF
--- a/app/tests/services/health.tests.js
+++ b/app/tests/services/health.tests.js
@@ -1,0 +1,49 @@
+describe('Health', function() {
+  'use strict';
+
+  var mockHealthService;
+
+  beforeEach(module('common'));
+  beforeEach(module('crate'));
+  beforeEach(module('health'));
+  beforeEach(function() {
+    angular.mock.inject(function($injector) {
+      mockHealthService = $injector.get('Health');
+    });
+  });
+
+  describe('mockHealthService', function() {
+
+    it('should return 0', inject(function() {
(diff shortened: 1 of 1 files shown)


app/tests/services/health.tests.js
```diff
@@ -0,0 +1,50 @@
...
+        shards_missing: 0,
+        shards_started: 18,
+        shards_underreplicated: 0,
+        size: 187297,
+        summary: "6 Shards / 0-all Replicas",
+        table_schema: "doc",
+        type_display_name: "TABLE.RECORDS",
+      };
+
+      var HEALTH = mockHealthService.fromString(tableInfo.health);
```
chaudum: no need to define a full tableinfo (it's just confusing)

```js
var health = mockHealthService.fromString('good');
expect(health.name).toEqual('good');
expect(health.level).toEqual(0);
```

m-kharbat: @chaudum removed tableinfo

app/tests/services/health.tests.js
```diff
@@ -0,0 +1,26 @@
...
+    });
+  });
+
+  describe('mockHealthService', function() {
+
+    it('should create health object from tableInfo health', inject(function() {
+      var health = mockHealthService.fromString('good');
+      expect(health.name).toEqual("good");
+      expect(health.level).toEqual(0);
+
```
chaudum: also add `warning` and `critical`?

PR #321, opened 2016-11-28
added tests for nodeinfo.js

m-kharbat:
@chaudum please review


app/tests/services/nodeInfo.nodeinfo.test.js
```diff
@@ -0,0 +1,417 @@
+describe('NodeInfo Services', function() {
```
chaudum: hm, not sure about this test. IMO it would be better to test the `queryResultToObjects` factory
https://github.com/crate/crate-admin/blob/master/app/scripts/services/sql.js#L4-L19

app/scripts/services/sql.js
```diff
@@ -12,6 +12,9 @@ angular.module('sql', [])
       return obj;
     };
     return function(sqlQuery, headers) {
```
chaudum: suggestion:

```js
  .factory('queryResultToObjects', function() {
    var toObject = function(cols, row) {
      var obj = {};
      if (cols.length === row.length) {
        for (var i = 0; i < cols.length; i++) {
          obj[cols[i]] = row[i];
        }
      }
      return obj;
    };
    return function(sqlQuery) {
      return sqlQuery.rows.map(function(o) {
        return toObject(sqlQuery.cols, o);
      });
    };
})
```

this however would require to name the select list items correctly.

PR #332, opened 2016-12-08
Lowered opacity of placeholder query in console

Autophagy:


Autophagy:
Old placeholder (same white as a regular query, so it looks like its actually in the console):
old-placeholder

New placeholder (half opacity, but regular queries are same opacity):
new-placeholder

PR #336, opened 2016-12-12
Removed pepper widget and moved support to help section.

Autophagy:
@celaus Could you check and make sure the order & wording of these is to your liking? Thanks!

image


m-kharbat:
is it intented that the get started is the last in the menu ? 
screen shot 2016-12-12 at 11 27 50

Autophagy:
Yep! Since its a help/support section we felt it would be best at the bottom.

PR #340, opened 2016-12-13
Admin UI console results visual improvements

Autophagy:


app/scripts/controllers/console.js
```diff
@@ -79,6 +84,43 @@ angular.module('console', ['sql'])
           });
         };
 
+        $scope.isGeopoint = function isGeopoint(dataType) {
```
m-kharbat: All these checking functions can be moved to a separate service which will make them reusable and a lot easier to test

app/scripts/controllers/console.js
```diff
@@ -232,5 +281,107 @@ angular.module('console', ['sql'])
...
+  .directive('formattedArray', function ($filter) {
+    return {
+      restrict: 'E',
+      scope: {
+        array: '=',
+        typesarray: '=',
+        expand: '='
+      },
+      templateUrl: 'views/formatted-array.html',
+      link: function($scope) {
```
m-kharbat: it's a bit confusing to have `$scope` injected in the link function, the standard way is to use `scope` https://www.bennadel.com/blog/2716-when-to-use-scope-vs-scope-in-angularjs.htm

app/scripts/controllers/console.js
```diff
@@ -232,5 +281,107 @@ angular.module('console', ['sql'])
...
+          } else {
+            return $filter('columnTypeClass')($scope.typesarray[1]);
+          }
+        };
+
+        $scope.getArrayType = function getArrayType() {
+          return $scope.typesarray[0] == 101 ? 'Set' : 'Array';
+        };
+
+        $scope.isObject = function isObject(index) {
```
m-kharbat: these functions are defined twice.. you can put them in a service and reuse them in the directive, which will also be easier to test

app/scripts/controllers/console.js
```diff
@@ -232,5 +281,107 @@ angular.module('console', ['sql'])
...
+  })
+  .directive('formattedObject', function () {
+    return {
+      restrict: 'E',
+      scope: {
+        object: '=',
+        expand: '='
+      },
+      templateUrl: 'views/formatted-object.html',
+      link: function($scope) {
```
m-kharbat: same comment about $scope in link function

app/styles/dark-theme-variables.less
```diff
@@ -254,3 +255,12 @@
 
 @code-bg-color: @lightred;
 @code-color:    @darkred;
+
+//
+// Console Results Colourisation
+// ---------------------------------------------------
+
+@numeric-result-color: #a0a1ff;
```
m-kharbat: all the colors , appart from the grays :( are defined in the App Colors section with color names to make them reusable and easily identifiable, also so that we don't have too many shades of the same color 

https://github.com/crate/crate-admin/blob/master/app/styles/light-theme-variables.less#L33

app/styles/light-theme-variables.less
```diff
@@ -255,3 +256,12 @@
 
 @code-bg-color:     @lightred;
 @code-color:        @darkred;
+
+//
+// Console Results Colourisation
+// ---------------------------------------------------
+
+@numeric-result-color: #181bd0;
```
m-kharbat: same comment about the colors

app/tests/filter/numbers.js
```diff
@@ -0,0 +1,122 @@
+describe('floor', function() {
```
m-kharbat: these tests are not related to the feature, you can add them here https://github.com/crate/crate-admin/pull/321/files or make a separate pr

app/views/formatted-array.html
```diff
@@ -0,0 +1,13 @@
+<div class="formatted-array">
```
m-kharbat: maybe we can add a suffix to know which files are directive templates and which are views ?

app/views/formatted-array.html
```diff
@@ -0,0 +1,13 @@
...
+  <div class="formatted-array__title"><a ng-click="toggleExpand()"><i class="fa fa-chevron-right" ng-class="isExpanded ? 'expanded-icon' : ''"></i> {{ getArrayType() }}<span class="formatted-array__length">[{{ getArrayLength() }}]</span></a></div>
+  <div class="formatted-array__elements" ng-if="isExpanded">
+    <div ng-repeat="element in array track by $index">
+      <span ng-if="isObject($index)" class="formatted-array__element">
+        <div class="key">{{ formatArrayindex($index) }}</div> <div class="seperator">:</div> <formatted-object object="element" expand="false"></formatted-object>
```
m-kharbat: i didn't see where some classes are defined example, key, value, separator etc

app/views/formatted-object.html
```diff
@@ -0,0 +1,16 @@
+<div class="formatted-object">
+  <div class="formatted-object__title"><a ng-click="toggleExpand()"><i class="fa fa-chevron-right" ng-class="isExpanded ? 'expanded-icon' : ''"></i> Object</a></div>
```
m-kharbat: how do you feel about using http://fontawesome.io/icon/caret-right/ (because we also using in the nodes view )?

app/scripts/controllers/console.js
```diff
@@ -232,5 +281,107 @@ angular.module('console', ['sql'])
       }
     };
   })
+  .directive('lazyLoadScroll', function ($window) {
+    return {
+      restrict: 'A',
+      link: function (scope, element, attrs) {
+        var e = element[0];
+        $($window).bind('scroll', function () {
```
m-kharbat: this looks a bit weird because you are injecting a $window to use as a jquery selector.. you can use `window` directly as a selector in jquery and you already have access to .scroll() event handler https://api.jquery.com/scroll/ it looks like .bind() is deprecated

app/views/formatted-array.html
Autophagy: Similarly, do you think the formatted-array/object templates should be moved into their own folder? app/views/templates or app/templates or something like this?

Autophagy:
@m-kharbat I think that's all your comments addressed apart from https://github.com/crate/crate-admin/pull/340#discussion_r92606960 - would be interesting to know your thoughts on it & whether it would involve refactoring other templates like https://github.com/crate/crate-admin/blob/mika/consoleResultsImprovements/app/views/editor.html

app/views/formatted-array.html
m-kharbat: I would suggest to just add `-template` or `-directive`
I think changing the folder would require a lot of change in the grunt build

PR #341, opened 2016-12-15
Fixed paragraph width on large screens

Autophagy:


app/styles/common.less
```diff
@@ -9,6 +9,10 @@ body {
   font-size: @font-size-small;
 }
 
+p {
```
m-kharbat: you are giving all the paragraphs a 100% width.

this issue occurs because there is a display flex on the container with flex-direction = row and flex-wrap = wrap. You should make the flex-direction = column and no wrap because there is no need for it.

or just wrap the elements inside in a div with display = block

PR #346, opened 2017-01-11
preparing release 1.1.0

m-kharbat:


m-kharbat:
the versions are correct in 1.1 https://github.com/crate/crate-admin/blob/1.1/bower.json#L4 https://github.com/crate/crate-admin/blob/1.1/package.json#L4

PR #350, opened 2017-01-13
show loading indicator when execute query is in progress

m-kharbat:
buf6ixbzqc

### DIFF
--- a/app/views/console.html
+++ b/app/views/console.html
@@ -42,7 +42,10 @@
 
             <div class="cr-console-header__options">
               <div class="cr-console-header__options__item--hint">{{ 'CONSOLE.HINT' | translate }}</div>
-              <div id="execute-btn" class="cr-console-header__options__item--btn cr-button cr-button--primary" ng-click="execute()" type="submit" class="ladda-button" data-style="expand-left" translate>{{ 'CONSOLE.EXCUTE_QUERY' }}</div>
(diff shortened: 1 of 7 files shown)


app/views/console.html
```diff
@@ -42,7 +42,10 @@
 
             <div class="cr-console-header__options">
               <div class="cr-console-header__options__item--hint">{{ 'CONSOLE.HINT' | translate }}</div>
-              <div id="execute-btn" class="cr-console-header__options__item--btn cr-button cr-button--primary" ng-click="execute()" type="submit" class="ladda-button" data-style="expand-left" translate>{{ 'CONSOLE.EXCUTE_QUERY' }}</div>
+              <div id="execute-btn" class="cr-console-header__options__item--btn cr-button cr-button--console" ng-click="execute()" type="submit" >
+                <div ng-class="loading ? 'hidden' : ''" translate>{{ 'CONSOLE.EXCUTE_QUERY' }}</div>
```
chaudum: could you replace `ng-class="loading ? 'hidden' : ''"` with `ng-if="!loading"`?

chaudum: or is it for keeping the width?

m-kharbat: yes it's for the width

chaudum:
all fixes also need to be backported to 1.0, since 1.1 has not been released (together with crate) yet.

PR #358, opened 2017-01-19
update charting library to angular-nvd3

m-kharbat:
i updated the charting library to https://github.com/krispo/angular-nvd3 which is better maintained and uses a later version of nvd3

http://g.recordit.co/xH6ygH5EoE.gif

### DIFF
--- a/app/views/overview.html
+++ b/app/views/overview.html
@@ -68,7 +68,7 @@
           <span class="cr-radio-button__load"><span id="load-btn-2" class="cr-radio-button__load__toggle cr-radio-button__load__toggle--fifteen"></span></span>
           Load 15</a>
       </div>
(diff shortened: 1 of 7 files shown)


app/views/overview.html
```diff
@@ -68,19 +68,7 @@
...
-                       yaxisshowmaxmin="false"
-                       objectequality="true"
-                       forcey="[0,1]"
-                       transitionduration="0"
-                       interactive="false"
-                       margin="{top: 20, right: 2, bottom: 40, left: 40}"
-                       showLegend="false">
-        <svg></svg>
-      </nvd3-line-chart>
+      <nvd3 id="cluster-load" on-ready='callback' api='api' options='options' data='chart.data'></nvd3>
```
chaudum: please use double quotes `"` around attribute values

PR #364, opened 2017-01-26
Added external link for formatted geoshape results

Autophagy:
geoshape

### DIFF
--- a/app/scripts/filter/text.js
+++ b/app/scripts/filter/text.js
@@ -120,4 +120,9 @@ angular.module('common')
     return function(timestamp) {
       return new Date(timestamp).toUTCString();
     };
+  })
+  .filter('urlEncodedJson', function() {
+    return function(json) {
+      return encodeURIComponent(JSON.stringify(json));
+    };
   });
--- a/app/views/console.html
+++ b/app/views/console.html
@@ -85,11 +85,11 @@
                     <pre class="query-result {{ resultHeaderTypes[$index] | columnTypeClass }}">{{ column }}<span  class="unsafe-number" data-original-title="{{ 'CONSOLE.UNSAFE_INTEGER' | translate }}" cr-tooltip cr-tooltip-position="right" ng-if="!ColumnTypeCheck.isSafe(column)"><i class="fa fa-exclamation-triangle" aria-hidden="true"></i></span></pre>
                   </div>
                   <div ng-if="column != null && !ColumnTypeCheck.requiresNoFormatting(resultHeaderTypes[$index])">
-                    <pre class="query-result {{ resultHeaderTypes[$index] | columnTypeClass }}" ng-if="ColumnTypeCheck.isGeopoint(resultHeaderTypes[$index])">{{ column }} <a href="https://www.google.com/maps/place/{{column[1]}},{{column[0]}}" target="_blank"><i class="fa fa-external-link"></i></a></pre>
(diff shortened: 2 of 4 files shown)


Autophagy:
@chaudum This links to http://geojson.io/ - but the `geo_point` data type links to a google maps link. Do you think this is fine, or do you think both should use geojson, or at least openmapbox?

chaudum:
@Autophagy I like geojson.io because you can also see the geojson input next to the map. I would point both types to geojson.io

app/views/console.html
```diff
@@ -89,7 +89,7 @@ <h2  class="query-result-container__header">{{ 'CONSOLE.RESULT_FROM_QUERY' | tra
...
                     <pre class="query-result {{ resultHeaderTypes[$index] | columnTypeClass }}" ng-if="ColumnTypeCheck.requiresArrayFormatting(resultHeaderTypes[$index])"><formatted-array array="column" typesarray="resultHeaderTypes[$index]" expand="true"></formatted-array></pre>
                     <pre class="query-result {{ resultHeaderTypes[$index] | columnTypeClass }}" ng-if="ColumnTypeCheck.requiresJSONFormatting(resultHeaderTypes[$index])"><formatted-object object="column" expand="true"></formatted-object></pre>
-                    <pre class="query-result {{ resultHeaderTypes[$index] | columnTypeClass }}" ng-if="ColumnTypeCheck.isGeoarea(resultHeaderTypes[$index])"><span>Type</span>: {{ column['type']}}<br/><span>Coordinates</span>: {{ column['coordinates'] }}</pre>
+                    <pre class="query-result {{ resultHeaderTypes[$index] | columnTypeClass }}" ng-if="ColumnTypeCheck.isGeoarea(resultHeaderTypes[$index])"><span>Type</span>: {{ column['type']}} <a href="http://geojson.io/#data=data:application/json,{{ column | formatGeoshape }}" target="_blank"><i class="fa fa-external-link"></i></a><br/><span>Coordinates</span>: {{ column['coordinates'] }} </pre>
```
m-kharbat: you should use ng-href instead of href:
https://docs.angularjs.org/api/ng/directive/ngHref

app/scripts/filter/text.js
```diff
@@ -120,4 +120,9 @@ angular.module('common')
     return function(timestamp) {
       return new Date(timestamp).toUTCString();
     };
+  })
+  .filter('formatGeoshape', function() {
```
chaudum: filter name and function implementation do not match.
I would name the filter `urlEncodedJson` (or similar)

if it is not too expensive calling multiple filters on an object I would even make two filters out of it, one for `JSON.stringify` and one for `encodeURIComponent`

Autophagy:
@chaudum I can point a geopoints coordinates at geojson.io but it doesnt place a marker down at the point, like gmaps does - it just centers the map on that location. A solution could be to encode the lat & long of a geopoint into a geoarea point, like:

```
{
        "type": "Point",
        "coordinates": [
          15.380859374999998,
          47.81315451752768
        ]
}```

Which would place a marker, but I worry it would be confusing. OSM does support markers in the URL based on lat/long, like this http://www.openstreetmap.org/?mlat=51.510772705078125&mlon=0.054931640625#map=9/51.5108/0.0549

chaudum:
@Autophagy so your suggestion would be to use OSM for both geo_point and geo_shape?

Autophagy:
@chaudum My suggestion would be to use geojson.io for geo_shape and OSM for geo_point. Geojson *uses* OSM, so both maps would be at least visually consistent

PR #367, opened 2017-02-01
Added a CrateDB mimetype, instead of using plsql

Autophagy:
partitioned

### DIFF
--- a/app/tests/controllers/console.test.js
+++ b/app/tests/controllers/console.test.js
@@ -1,0 +1,27 @@
+describe('Console', function() {
+  'use strict';
+
+  var KeywordObjectCreator;
+
+  beforeEach(module('crate'));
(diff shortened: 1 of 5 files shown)


app/tests/controllers/console.test.js
```diff
@@ -0,0 +1,21 @@
...
+  });
+
+  describe('KeywordObjectCreator', function() {
+    it('should create object from keyword array', inject(function() {
+      var keywordArray = ['select', 'from', 'limit'];
+      var keywordsObject = KeywordObjectCreator.create(keywordArray);
+      expect(keywordsObject).toEqual({'select': true, 'from': true, 'limit': true});
+    }));
+  });
+
```
chaudum: optional: add another test that checks what happens if you have duplicate keywords in the initial array

PR #370, opened 2017-02-06
Added buttons to collapse and expand all schemas in the tables view

m-kharbat:
http://g.recordit.co/qHdvyZpbgT.gif


m-kharbat:
the best way would be to remove the text http://recordit.co/hbbvGqZOc8

PR #380, opened 2017-02-27
Expanded the list of CrateDB keywords in console

Autophagy:


m-kharbat:
do we need a changelog for this ?

Autophagy:
@m-kharbat I don't think so, since the feature is currently unreleased and this is just a fix/update for that: https://github.com/crate/crate-admin/blob/master/CHANGES.txt#L18

Also, the CI fails because of:
```
if(SettingsProvider.$get().enterprise == true){
  loadStylesheet('static/styles/main-enterprise.css');
}
````
Shall I also change this to `=== true` in this PR?

PR #387, opened 2017-03-06
remove unused dependencies

m-kharbat:
removed unused dependency from the admin-ui


app/index.html
```diff
@@ -27,27 +27,23 @@
     <!-- endbuild -->
 
     <!-- build:js static/scripts/angular.js -->
-    <script src="bower_components/angular/angular.js"></script>
-    <script src="bower_components/angular-route/angular-route.js"></script>
-    <script src="bower_components/angular-cookies/angular-cookies.js"></script>
-    <script src="bower_components/angular-truncate-2/src/angular-truncate-2.js"></script>
-    <script src="bower_components/angular-sanitize/angular-sanitize.js"></script>
+    <script src="bower_components/angular/angular.min.js"></script>
```
chaudum: I would not include minimised versions here, because it is harder to debug and the files are concatenated and minified in the build process anyway.

chaudum: not sure how minifying minified javascript works out ...

m-kharbat: I assumed the minified files would be concatenated, because it worked properly when I tested  after building the app,  but I'm not sure. 

I will remove the min dependencies. But I think it's best to use the official minified versions, i will work on a solution in the free sprint.

PR #395, opened 2017-03-17
Removed odd behaviour with load readings on Windows

Autophagy:
@m-kharbat Do you think this needs to be backported?


m-kharbat:
yes, it should be backported to 1.2

PR #399, opened 2017-03-23
add fallback value for nodeInfo to avoid js errors

m-kharbat:
error


m-kharbat:
@chaudum it's probably some other reason crate on docker doesn't have this stats problem

PR #400, opened 2017-03-23
display notification only when a crate version is out

m-kharbat:


app/views/statusbar.html
```diff
@@ -34,8 +34,8 @@
     <div ng-controller="NotificationsController">
       <div id="settings-icon" class="status-bar__content__item" ng-click="toggleSettings()" data-original-title="Settings" cr-tooltip cr-tooltip-position="bottom">
         <i class="fa fa-2x fa-cog"></i>
-        <i id="unread-notifications-bg" ng-if="numUnread || showUpdate" class="fa fa-circle"></i>
-        <i id="unread-notifications" ng-if="numUnread || showUpdate" class="fa fa-exclamation-circle"></i>
+        <i id="unread-notifications-bg" ng-if="showUpdate" class="fa fa-circle"></i>
```
Autophagy: This is the only place `numUnread` is used. If we are removing it, perhaps it would make sense to remove it entirely from the feed controller? https://github.com/crate/crate-admin/blob/2ccb9fbd76e81b86e291a3c170ac0edf4f7fdf65/app/scripts/controllers/feed.js#L67

m-kharbat: the whole blog feed will be removed in the next prs

PR #419, opened 2017-04-11
add screenshots

nslater:
I did this because I am adding screenshots to the crate/crate README, but we need to keep them up to date. so I've shuffled things around so the screenshots are getting updated during the crate-admin release process (but only when necessary)

I've also added a thing to the crate-admin release process that updates the screenshot used in the crate/crate README

this isn't perfect, because it the admin-ui version in crate/crate is updated separately. so there may be times when the two things are out of sync

but I couldn't figure out how to get the two things in sync. as there's no standard process to follow to update the admin-ui version in crate/crate

(I tried to make it so the screenshots were bundled as part of the crate-admin distribution, but half way through that I realised that the admin-ui is fetched at build time, not bundled in the source. so that's useless as this screenshot is needed for the readme on github only)


nslater:
closing this

we will solve this another way

PR #422, opened 2017-04-13
display license.ident in the admin-ui

m-kharbat:
screen shot 2017-04-13 at 16 47 38
screen shot 2017-04-13 at 16 48 12
screen shot 2017-04-13 at 16 48 17

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,6 +5,17 @@ Changes for Crate Admin Interface
 Unreleased
 ==========
 
+Breaking Changes
+================
+
+Changes
+=======
+
+ - License ident is now displayed in the Enterprise Edition.
+
+Fixes
+=====
+
 2017/04/12 1.2.3
 ================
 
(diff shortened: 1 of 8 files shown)


CHANGES.txt
```diff
@@ -5,6 +5,8 @@ Changes for Crate Admin Interface
 Unreleased
 ==========
 
+ - Display ``license.ident`` in the Enterprise Edition.
```
Autophagy: Are we making the changelog messages more user focused/friendly? If so, then I would go with something like 'License ident is now displayed in the Enterprise Edition".

Autophagy: Also, since this is a thing we're now doing for the next release (but not backporting it), can you add the following sections to the CHANGE.txt under unreleased:

```
Breaking Changes
================

Changes
=======

Fixes
====
```

And to put this in the corresponding section (`Changes` I assume?)

PR #432, opened 2017-07-05
redirect to /401 when the admin-ui user is not allowed to use the cluster

m-kharbat:
redirect to /401, if the user has no cluster privilege, but not if the console query fails with 401
screen shot 2017-07-05 at 15 48 09


app/scripts/services/sql.js
```diff
@@ -92,6 +92,8 @@ angular.module('sql', [])
         data.args = args;
       }
 
+      isConsole = typeof isConsole !== 'undefined' ? isConsole : false;
```
chaudum: I know it's a pain in the ass to change all method calls to use the new signature, but imo the typeof check is not best practice. it also has the flaw, that you could pass any type that is not boolean which would break the `not` check

app/views/401.html
```diff
@@ -0,0 +1,13 @@
...
+		<img class="unauthorized-header__icon" src="static/assets/logo-CrateDB@2x.png">
+	</div>
+
+	<div class="unauthorized-body">
+	 <img class="unauthorized-body__img" src="static/assets/gfx-not-allowed@2x.png">
+	 <h3>You are not allowed to access this cluster</h3>
+	 <h4>Please contact your system administrator to get more information</h4>
+	</div>
+</div>
+</div>
```
m-kharbat: new line character added

app/scripts/services/sql.js
```diff
@@ -83,15 +83,15 @@ angular.module('sql', [])
     };
 
 
-    SQLQuery.execute = function(stmt, args, errorTrace, getTypes) {
+    SQLQuery.execute = function(stmt, args, errorTrace, getTypes, isConsole) {
       var data = {
         'stmt': stmt
       };
 
       if (args !== undefined && !$.isEmptyObject(args)) {
```
chaudum: args cannot be undefined any more

app/styles/common.less
```diff
@@ -537,3 +537,57 @@ label:hover:before {
...
+  height: 500px;
+
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+}
+.unauthorized-body__img {
+  width: 50%;
+}
```
chaudum: newline character missing - please check if your editor can automatically add newlines

app/scripts/services/sql.js
```diff
@@ -139,17 +139,22 @@ angular.module('sql', [])
...
           }
+
           deferred.reject(new SQLQuery(stmt, data, error));
         });
       return promise;
     };
 
     return SQLQuery;
   });
+
```
chaudum: newline character missing - please check if your editor can automatically add newlines

CHANGES.txt
```diff
@@ -14,6 +14,8 @@ Changes
 Fixes
 -----
 
+ - Redirect to /401 when the admin-ui user doesn't have cluster privileges.
```
chaudum: ... does not have permission to access the cluster.

PR #445, opened 2017-07-17
Fix insert statement for importing tweets

mxm:
The insert statements for the Twitter tutorial does not specify the
columns names. The order of the columns somehow changed and the insert
statement does not work anymore. Specifying the columns explicitly fixes
the problem.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -14,6 +14,8 @@ Changes
 Fixes
 -----
 
+ - Fix Tweet import for Twitter tutorial page
+
 2017/07/11 1.4.1
 ================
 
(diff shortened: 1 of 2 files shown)


CHANGES.txt
```diff
@@ -14,6 +14,8 @@ Changes
 Fixes
 -----
 
+ - Fix Tweet import
```
chaudum: A bit more verbose, e.g. "Fixed tweet import in tutorial view."

mxm:
Merged with a5da72592a87f81e632b10dbb9d6e6bb94b8f9a9.

PR #446, opened 2017-07-24
Changed documentation link to x.y link instead of x.y.z

Autophagy:
This is due to our decision to shift away from building seperate documentation for each hotfix release, and to instead only build for feature releases.

### DIFF
--- a/app/scripts/controllers/common.js
+++ b/app/scripts/controllers/common.js
@@ -48,8 +48,12 @@ var commons = angular.module('common', ['stats', 'udc'])
     }, {});
 
     var DOCS_BASE_URL = 'https://crate.io/docs';
+    var getMajorMinorVersion = function getMajorMinorVersion(version) {
+      var SHORT_VERSION_RE = new RegExp(/(\d+\.\d+)\.\d+.*/);
+      return version ? version.number.match(SHORT_VERSION_RE)[1] : null;
+    };
     var getDocsUrl = function getDocsUrl(version) {
-      return $sce.trustAsResourceUrl(DOCS_BASE_URL + (version ? '/en/' + version.number : '/stable') + '/');
+      return $sce.trustAsResourceUrl(DOCS_BASE_URL + (version ? '/en/' + version : '/stable') + '/');
     };
     $scope.cluster_color_label = '';
     $scope.config_label = '';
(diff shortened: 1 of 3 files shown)


app/scripts/controllers/common.js
```diff
@@ -48,8 +48,12 @@ var commons = angular.module('common', ['stats', 'udc'])
     }, {});
 
     var DOCS_BASE_URL = 'https://crate.io/docs';
+    var getMajorMinorVersion = function getMajorMinorVersion(version) {
+      var splitVersion = version.split('.');
+      return splitVersion[0] + '.' + splitVersion[1];
+    };
     var getDocsUrl = function getDocsUrl(version) {
-      return $sce.trustAsResourceUrl(DOCS_BASE_URL + (version ? '/en/' + version.number : '/stable') + '/');
+      return $sce.trustAsResourceUrl(DOCS_BASE_URL + (version ? '/en/' + version : '/stable') + '/');
```
turbo-ele: i think the default shouldn't point to `stable`, since we removed that

app/scripts/controllers/common.js
```diff
@@ -123,7 +127,8 @@ var commons = angular.module('common', ['stats', 'udc'])
       $scope.load15 = data.load[2] == '-.-' ? data.load[2] : data.load[2].toFixed(2);
       $scope.load15 = $scope.load15 < 0 ? 'N/A' : $scope.load15;
       $scope.version = data.version;
-      $scope.docs_url = getDocsUrl(data.version);
+      $scope.major_minor_version = getMajorMinorVersion(data.version.number);
```
turbo-ele: what if data.version is null?

app/scripts/controllers/common.js
```diff
@@ -48,8 +48,12 @@ var commons = angular.module('common', ['stats', 'udc'])
     }, {});
 
     var DOCS_BASE_URL = 'https://crate.io/docs';
+    var getMajorMinorVersion = function getMajorMinorVersion(version) {
+      var splitVersion = version.split('.');
```
turbo-ele: alternatively you could use a regex as so `/\d+\.\d+/`

chaudum:
```js
var SHORT_VERSION_RE = new RegExp(/(\d+\.\d+)\.\d+.*/);
var shortVersion = version.match(SHORT_VERSION_RE)[1];
```

PR #463, opened 2017-09-13
Added query table button to table view

m-kharbat:
http://g.recordit.co/rYNat3jjAJ.gif

### DIFF
--- a/app/scripts/controllers/console.js
+++ b/app/scripts/controllers/console.js
@@ -14,7 +14,7 @@ angular.module('console', ['sql', 'datatypechecks'])
   .directive('console', function(SQLQuery, ColumnTypeCheck){
     return {
       restrict: 'A',
-      controller: ['$scope', '$translate', '$location', 'Clipboard', function($scope, $translate, $location, Clipboard){
+      controller: ['$scope', '$translate', '$location', 'Clipboard', '$timeout', function($scope, $translate, $location, Clipboard, $timeout){
         var self = this;
 
         var inputScope = null;
@@ -67,7 +67,9 @@ angular.module('console', ['sql', 'datatypechecks'])
           if ($scope.useLocalStorage) {
             localStorage.setItem('crate.console.query_list', JSON.stringify(self.recentQueries));
           }
-          inputScope.recentCursor = -1;
+          if (inputScope) {
+            inputScope.recentCursor = -1;
+          }
         };
 
         $scope.hide = function hide(item){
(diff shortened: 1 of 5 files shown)


app/scripts/controllers/tables.js
```diff
@@ -106,6 +111,29 @@ angular.module('tables', ['stats', 'sql', 'common', 'tableinfo'])
       $scope.renderSiderbar = true;
       $scope.isParted = false;
 
+      var constructQuery = function (rows) {
+        var result = 'SELECT ';
```
chaudum: maybe rename result to `stmt` or `query` for better readability

app/scripts/controllers/tables.js
```diff
@@ -106,6 +111,29 @@ angular.module('tables', ['stats', 'sql', 'common', 'tableinfo'])
...
+        var current_object;
+
+        for (var i = 0; i < rows.length; i++) {
+          if (current_object !== '' && rows[i].data_type === 'object' && 
+            !rows[i].column_name.startsWith(current_object)) {
+            current_object = rows[i].column_name;
+            result += '\"' + rows[i].column_name + '\", ';
+          }
+        
+          if (current_object && rows[i].column_name.startsWith(current_object)) {
```
chaudum: I think this logic does not work correctly. imagine a table with the following schema:

```sql
CREATE TABLE t1 (a OBJECT, ab OBJECT, abc OBJECT);
```

app/scripts/controllers/tables.js
```diff
@@ -106,6 +111,29 @@ angular.module('tables', ['stats', 'sql', 'common', 'tableinfo'])
...
+            //in the case of an object column we don't want to get the nested objects
+            continue;
+          } else {
+            result += '\"' + rows[i].column_name + '\", ';
+          }
+        }
+        result = result.substring(0, result.length - 2);
+        result += ' FROM ' + tableSchema + '.' + tableName + ' LIMIT 100;';
```
chaudum: tableSchema and tableName need to be double quoted

app/scripts/controllers/console.js
```diff
@@ -67,7 +67,9 @@ angular.module('console', ['sql', 'datatypechecks'])
           if ($scope.useLocalStorage) {
             localStorage.setItem('crate.console.query_list', JSON.stringify(self.recentQueries));
           }
-          inputScope.recentCursor = -1;
+          if (inputScope) {
```
chaudum: what if `inputScope` is `0`?

```js
if (inputScope !== null) {
    inputScope.recentCursor = -1;
}
```

m-kharbat: the issue is when `inputScope` is null, i get a referenceError.
in line 359 we set it to `$console.setInputScope($scope);` I'm not sure when it can be 0

PR #471, opened 2017-09-22
improve ng-repeat performance

m-kharbat:
improved performance for the shards view from 5,284 watchers (master) to 1,432 watchers (current branch). Tested with data from the staging cluster.

Instead of having one custom tooltip per shard-item and per node-item, and a mouseenter/mouseleave for each.. we now have only one tooltip that updates it's own position. There is only 2 listeners on the tbody element. These listeners, trigger the update of the tooltip position, and data.

The tooltip is a transclude directive (https://docs.angularjs.org/api/ng/directive/ngTransclude) so the nodeContent/shardContent directives were removed for better performance.


app/plugins/shards/shards.html
```diff
@@ -2,22 +2,22 @@
   <div class="page-container">
       <h1 class="cr-page-header">Shards</h1>  
 
-      <div class="cr-panel-block" ng-if="empty">
+      <div class="cr-panel-block" ng-show="empty">
```
chaudum: should this be the exact opposite of the other block?
`ng-show="empty || !shards"`

I know this is kind of unrelated. maybe in a separate commit?

app/plugins/shards/shards.js
```diff
@@ -8,14 +8,14 @@ angular.module('shards', ['sql'])
 
     var stmt = 'SELECT shards.id, size, min_lucene_version, num_docs, orphan_partition,  shards.table_name, shards.schema_name, shards.partition_ident, format(\'%s.%s\', shards.schema_name, shards.table_name) AS fqn, ' +
       'format(\'%s.%s.%s\', shards.schema_name, shards.table_name, shards.partition_ident) AS key, ' +
-      '_node[\'id\'] AS node_id, _node AS nodeInfo, state, routing_state, relocating_node, primary, ' +
+      '_node[\'id\'] AS node_id, _node[\'name\'] AS nodeInfo_name, _node[\'hostname\'] AS nodeInfo_hostname, _node[\'rest_url\'] AS nodeInfo_rest_url, _node[\'version\'][\'number\'] AS nodeInfo_version, state, routing_state, relocating_node, primary, ' +
```
chaudum: I would not mix camelCase_and_snake_case ;)

app/plugins/shards/shards.js
```diff
@@ -105,21 +88,106 @@ angular.module('shards', ['sql'])
...
+    $('tbody').mousemove(function (event) {
+        safeApply($scope, function () {
+          if ($scope.displayTooltip) {
+            $scope.pageX = event.pageX;
+            $scope.pageY = event.pageY;
+          }
+        });
+      })
+      .mouseover(function (event) {
+        var target;
```
chaudum: please create jQuery instance only once:

```js
var target = $(event.target);
if (target.hasClass(...) && !target.parent().hasClass(...)) {
  target = target.parent();
} else if {
...
}
```

app/plugins/shards/shards.html
m-kharbat: i'll make a separate commit for it !

PR #473, opened 2017-10-03
WIP virtual scroll in console

chaudum:


app/styles/console.less
```diff
@@ -153,6 +153,11 @@ tr > .cr-table-cell {
   }
 }
 
+.vs-container {
+  max-height: 600px;
```
chaudum: the virtual scroll requires a fixed height of the container, however we want a variable height depending on the browser size. guess we need to do the resizing via JS @m-kharbat

chaudum:
outdated

PR #476, opened 2017-10-06
Removed automatic execution of table queries

m-kharbat:


CHANGES.txt
```diff
@@ -14,6 +14,8 @@ Changes
 Fixes
 -----
 
+ - Removed automatic execution of table queries.
```
chaudum: I think it is not very clear what table queries are. Maybe explain a bit more, e.g:

- Do not automatically execute the query when clicking on the "Query Table" button in the table view.

PR #478, opened 2017-10-06
add progress bar for shards view

m-kharbat:


app/plugins/shards/shards.html
```diff
@@ -2,10 +2,13 @@
   <div class="page-container">
       <h1 class="cr-page-header">Shards</h1>  
 
-      <div class="cr-panel-block" ng-show="empty || !shards">
+      <div class="cr-panel-block" ng-show="empty">
         There are no shards. Shards are created and started when you create tables.
       </div>
 
+      <div class="progress" ng-show="!empty && !shards">
+        <div class="indeterminate"></div>
```
Autophagy: Perhaps "loading" or something would be a better class to use than indeterminate - assuming that is what the progress bar is actually showing. Otherwise, looks cool!

m-kharbat: this is from material design terminology, it stands for indeterminate progress (eg, https://materialdoc.com/components/progress/#indeterminate)

PR #481, opened 2017-10-11
on demand loading of shard and node data

m-kharbat:
remove the one time binding in this view because it prevented from seeing the relocation

### DIFF
--- a/app/plugins/shards/shards.js
+++ b/app/plugins/shards/shards.js
@@ -1,22 +1,85 @@
 'use strict';
 
 angular.module('shards', ['sql'])
-  .factory('ShardService', function (SQLQuery, queryResultToObjects, $q) {
-    var ShardService = {
-      deferred: $q.defer()
-    };
-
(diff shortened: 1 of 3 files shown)


app/plugins/shards/shards.js
```diff
@@ -117,23 +188,44 @@ angular.module('shards', ['sql'])
       // set tooltip data
       var table_key = target.data('table');
       var node_id = target.data('node');
+      var shard_id = target.data('id');
+
+      var table_ids = table_key.split('.');
```
chaudum: table_ids is a confusing variable name
maybe `tableNameParts`?

app/plugins/shards/shards.js
```diff
@@ -117,23 +188,44 @@ angular.module('shards', ['sql'])
       // set tooltip data
       var table_key = target.data('table');
```
chaudum: please don't mix camelCase and snake_case

PR #492, opened 2017-10-18
redirect to the first node by default

m-kharbat:


CHANGES.txt
```diff
@@ -14,6 +14,9 @@ Changes
 Fixes
 -----
 
+ - fixed issue that caused the cluster view to not redirect
```
chaudum: Fixed

app/scripts/controllers/cluster.js
```diff
@@ -18,10 +18,16 @@ angular.module('cluster', ['stats', 'sql', 'common', 'nodeinfo'])
 
         var firstRedirect = false;
 
+        $scope.goToPath = function (id) {
```
chaudum: is this function also used in the HTML templates? if not, it does not be on $scope
you could call $state.go('nodes.node', {}) directly on line 30 then

PR #493, opened 2017-10-19
fixed /401 redirection

m-kharbat:
use correct state name for redirection to /401

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -14,6 +14,9 @@ Changes
 Fixes
 -----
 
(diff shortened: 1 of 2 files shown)


CHANGES.txt
```diff
@@ -14,6 +14,8 @@ Changes
 Fixes
 -----
 
+ - Fixed issue that caused the redirection to ``/401`` to fail.
```
chaudum: ... redirect to `/401` in case of unauthorized access to fail.

PR #498, opened 2017-10-24
Meri/cache shards

m-kharbat:


app/plugins/shards/shards.js
```diff
@@ -174,7 +176,7 @@ angular.module('shards', ['sql'])
       }
     }
 
-    function setShardData(target) {
+    function setShardData(target, event) {
```
chaudum: seems like `event` is never used?

app/plugins/shards/shards.js
```diff
@@ -201,7 +203,7 @@ angular.module('shards', ['sql'])
       $scope.displayTooltip = true;
     }
 
-    function setNodeData(target) {
+    function setNodeData(target, event) {
```
chaudum: same here

app/plugins/shards/shards.js
chaudum: ignore

PR #506, opened 2017-11-06
add initial calls to update table list

m-kharbat:
remove delay in displaying the tables list


Autophagy:
Oh, I almost forgot - please add a changelog entry

m-kharbat:
it was caused by a fix on master that wasn't released yet

Autophagy:
Ah! Ok, cool :slightly_smiling_face:

PR #510, opened 2017-11-07
fix issue that caused closed partitions to show critical data health

m-kharbat:
ignore closed partitions

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -14,6 +14,9 @@ Changes
 Fixes
 -----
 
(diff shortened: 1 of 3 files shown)


CHANGES.txt
```diff
@@ -14,6 +14,9 @@ Changes
 Fixes
 -----
 
+ - Fixed issue that caused closed partitions to show a 
+   `CRITICAL` data state.
```
chaudum: double backtick

Issue #515, opened 2017-11-23
There are no tables created yet


senchi:
I've installed latest version (v2.2.3) via `bash -c "$(curl -L try.crate.io)"` and created a simple table like:
```
create table test.table1 (column1 string, column2 string);
```

When I open the "table" tab in the Admin console I get the `There are no tables created yet.` message:

![image](https://user-images.githubusercontent.com/903970/33189181-b3ceac9e-d0a0-11e7-8324-fd417fd86efd.png)

This also happens when I turn OFF the Enterprise features!

My environment is: OSX v10.13.1 (macOS High Sierra), Intel Core i7

joemoe:
hi @senchi,

sorry to hear about your issues. 

* is there something in the log?
* can you query that table from the console?
* what is the output?
* what is the output when you query `select * From information_schema.tables limit 100;` 

thanks.

senchi:
> is there something in the log?

I can't see any movements in the log (I suppose because the Admin is cilent-side only?
I'm not sure if related, but when I click around in the Admin, there is this in the browser log:

![image](https://user-images.githubusercontent.com/903970/33262840-fb1ba38a-d367-11e7-9ebf-66218ce6952d.png)

> can you query that table from the console?

Yes.

> what is the output?

Well.. no rows because table is empty:

![image](https://user-images.githubusercontent.com/903970/33263158-0e114b2e-d369-11e7-95ba-b09c9fc703c0.png)

> what is the output when you query select * From information_schema.tables limit 100;

It seems like my table is there, well created and I can query it..
![image](https://user-images.githubusercontent.com/903970/33263251-652a34f2-d369-11e7-9bf8-a61c7503da8f.png)


Thanks,

joemoe:
ok @senchi. this looks like a frontend problem. thanks. this helps us.

m-kharbat:
@senchi something seems to be off with your cluster, you have Nodes: 0, Critical Data and Health stats and 0 Loads as well. And I wasn't able to reproduce it.

do you have any other significant errors in the browser's console, perhaps failed http queries ?

senchi:
Hi @m-kharbat , no, not really, at least I couldn't see any.. the `tables.html` gets loaded with a `200`.. 
What I did is clean install the Crate.IO via the install script:
```
$ bash -c "$(curl -L try.crate.io)"
```
No changes to the config file whatsoever.. 
When I turn off the enterprise feature, and restart Crate.IO the same happens.

Thanks

[...]

PR #516, opened 2017-11-27
refresh cluster info from console

m-kharbat:


m-kharbat:
@Autophagy the second commit fixes this issue https://github.com/crate/crate-admin/issues/515

PR #525, opened 2018-01-02
map the Tab key to insert spaces instead of a tab character

m-kharbat:


app/scripts/controllers/console.js
```diff
@@ -482,7 +482,19 @@ angular.module('console', ['sql', 'datatypechecks', 'stats'])
...
+          // indent using spaces
+          indentWithTabs: false,
+          indentUnit: 2,
+          tabSize: 2
+        });
+
+        // map the Tab key to insert spaces instead of a tab character
+        editor.setOption('extraKeys', {
+          Tab: function (cm) {
+            var spaces = Array(cm.getOption('indentUnit') + 1).join(' ');
```
chaudum: please use [String.prototype.repeat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/repeat)

PR #527, opened 2018-01-12
Fix an issue that caused the Cluster tab to not be loaded correctly

m-kharbat:


app/scripts/services/stats.js
```diff
@@ -119,16 +119,25 @@ angular.module('stats', ['sql', 'health', 'tableinfo', 'nodeinfo', 'events'])
...
+            var readBytesDelta = currentValue.data.bytes_read - lastValue.data.bytes_read;
+            var writtenBytesDelta = currentValue.data.bytes_written - lastValue.data.bytes_written;
+            node.iostats = {
+              'rps': currentValue.data.reads > 0 ? (readsDelta / timeDelta) : -1,
+              'wps':  currentValue.data.writes > 0 ? (writesDelta / timeDelta) : -1,
+              'rbps': currentValue.data.bytes_read > 0 ? (readBytesDelta / timeDelta) : -1,
+              'wbps': currentValue.data.bytes_written > 0 ? (writtenBytesDelta / timeDelta) : -1
+            };
+          } else {
+              node.iostats = {
```
chaudum: please create an EMPTY_STATS object that can be reused here and in line 142

app/scripts/services/stats.js
```diff
@@ -115,6 +115,13 @@ angular.module('stats', ['sql', 'health', 'tableinfo', 'nodeinfo', 'events'])
         if (node.fs) {
           currentValue.data= node.fs.total;
         }
+
+        var empty_iostats = {
```
chaudum: please define inside a scope that is only executed once

PR #533, opened 2018-02-12
implement authentication feature in the admin-ui

m-kharbat:
screen shot 2018-02-12 at 18 03 35
screen shot 2018-02-12 at 18 03 21

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -11,6 +11,8 @@ Breaking Changes
 Changes
 -------
 
+ - Implement Authentication feature in the admin ui. 
+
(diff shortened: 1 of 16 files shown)


app/views/auth.html
```diff
@@ -0,0 +1,3 @@
+<div>
```
Autophagy: What is the use of this template? :sweat_smile:

m-kharbat: @Autophagy authentication directive requires a template.. 🤔i need to find a better solution, for now i replaced with an inline template

PR #534, opened 2018-02-13
import docs from the getting started docs project

nslater:
imported docs from the getting started docs

I have copied this setup from the crate/crash repository

needs crate/crate-docs-theme/pull/111 to be merged and released before this will run

note: this is a single page. we could improve these docs by expanding them. but that is out of scope for the current sprint


nslater:
@chaudum pushed a fixup. please re-review. thanks!

nslater:
@chaudum oops. the screenshot was the wrong one. I have pushed a fixup with the right one

nslater:
thank you both!

PR #539, opened 2018-02-18
Updated the table list search to filter tables only by fqn

m-kharbat:
limited the search filter to the fqn and removed one-time binding


m-kharbat:
it was searching in all the table objects fields

PR #545, opened 2018-03-08
Update the Size label in the table detail to Size (sum of primary shards for this table)

m-kharbat:


app/views/table-detail.html
```diff
@@ -108,7 +108,7 @@ <h1>{{ 'NAVIGATION.TABLE' | translate }}</h1>
     <div class="table-detail__row">
       <div class="table-detail__cell">
         <div class="cr-panel-block__item__header">
-          {{ 'TABLE.SIZE' | translate }}
+          {{ 'TABLE.SIZE' | translate }} (Sum of primary shrads of this table)
```
Autophagy: ~~shrads~~ shards :wink: also I think __(Sum of primary shards)__ should suffice, its on the table view so that already suggests that its the primary shards of the table.

PR #552, opened 2018-03-28
updated information schema queries to return only tables

chaudum:


app/scripts/controllers/tables.js
```diff
@@ -251,8 +251,8 @@ angular.module('tables', ['stats', 'sql', 'common', 'tableinfo', 'events'])
             // Table Schema
             var tableStmt = 'SELECT column_name, data_type, is_generated, generation_expression ' +
               'FROM information_schema.columns ' +
-              'WHERE table_schema = ? AND table_name = ?';
-            SQLQuery.execute(tableStmt, [tableSchema, tableName], false, false, false, false)
+              'WHERE table_type = ? AND table_schema = ? AND table_name = ?';
```
chaudum: `table_type` does not exist in `information_schema.columns`

app/scripts/controllers/tables.js
```diff
@@ -251,7 +251,7 @@ angular.module('tables', ['stats', 'sql', 'common', 'tableinfo', 'events'])
             // Table Schema
             var tableStmt = 'SELECT column_name, data_type, is_generated, generation_expression ' +
               'FROM information_schema.columns ' +
-              'WHERE table_schema = ? AND table_name = ?';
+              'WHERE AND table_schema = ? AND table_name = ?';
```
chaudum: ~~AND~~

Autophagy: `WHERE AND`? Is something missing here? :confused:

chaudum: I forget to remove the AND because I had an additional filter before that was not necessary

chaudum: pushed fixup (same commit)

PR #553, opened 2018-03-29
fix up LICENCE and NOTICE

nslater:
following the official Apache guide:

http://www.apache.org/dev/licensing-howto.html

steps I took:

- started from two files:
  - LICENCE containing an unmodified version of the Apache License 2.0
  - NOTICE containing the standard Crate.io legal boilerplate
- added copyright and license information for "bits we're bundling" to the end of the LICENSE file
  - this meant moving some stuff from bootstrap.py (which is totally fine, and I left a note there, as is required by their licence)
  - it also meant dropping many of the things we were already listing, because, I poked around, and as far as I can tell, we are not directly "bundling those bits". we rely on the end user to fetch those bits when she is building the project. so there's no requirement for us to mention them in these files

this is just the first PR

I will not follow a similar process to fix up crate/crate-docs-theme and crate/crate

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,20 +1,3 @@
-Crate Admin Interface
-Copyright 2013-2015 CRATE Technology GmbH ("Crate")
-
-
-Licensed to CRATE Technology GmbH (referred to in this notice as "Crate")
-under one or more contributor license agreements.  See the NOTICE file
-distributed with this work for additional information regarding copyright
-ownership.
-
-Crate licenses this software to you under the Apache License, Version 2.0.
-However, if you have executed another commercial license agreement with
-Crate these terms will supersede the license and you may use the software
-solely pursuant to the terms of the relevant commercial agreement.
-
-
-=========================================================================
-
 
                                  Apache License
                            Version 2.0, January 2004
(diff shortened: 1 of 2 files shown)


bootstrap.py
```diff
@@ -1,16 +1,11 @@
-##############################################################################
+# Per the Zope Public License Version 2.0, you must update this file-specific
```
jodok: we only redistribute this file - this is standardized, so we should not modify it.

NOTICE
```diff
@@ -1,98 +1,13 @@
...
-Copyright (c) 2013 Christian Maurer
-License: Apache License, Version 2.0 (http://www.apache.org/licenses/LICENSE-2.0)
-
-=========================================================================
+Licensed to CRATE Technology GmbH (referred to in this notice as "Crate")
+under one or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information regarding copyright
+ownership.
 
+Crate licenses this software to you under the Apache License, Version 2.0.
```
jodok: i prefer to have this part as top of the LICENSE file - above the standardized apache license

NOTICE
```diff
@@ -1,98 +1,13 @@
-Crate Admin Interface
-Copyright 2013-2015 CRATE Technology GmbH
+CrateDB Admin Interface
+Copyright 2013-2018 CRATE Technology GmbH ("Crate")
```
jodok: the correct legal entity is Crate.IO GmbH

NOTICE
```diff
@@ -1,98 +1,13 @@
...
-
-=========================================================================
-
-angularjs-nvd3-directives
-http://angularjs-nvd3-directives.github.io/angularjs-nvd3-directives/
-Copyright (c) 2013 Christian Maurer
-License: Apache License, Version 2.0 (http://www.apache.org/licenses/LICENSE-2.0)
-
-=========================================================================
+Licensed to CRATE Technology GmbH (referred to in this notice as "Crate")
```
jodok: Crate.IO GmbH

LICENSE
```diff
@@ -1,20 +1,3 @@
-Crate Admin Interface
-Copyright 2013-2015 CRATE Technology GmbH ("Crate")
-
-
-Licensed to CRATE Technology GmbH (referred to in this notice as "Crate")
```
jodok: i prefer to keep this blurb above the license

LICENSE
```diff
@@ -217,3 +200,177 @@ solely pursuant to the terms of the relevant commercial agreement.
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
+
+================================================================================
```
jodok: i suggest a consistent delimiter. either 80 = or sth. like ```---``` and two newlines above?

LICENSE
```diff
@@ -217,3 +200,177 @@ solely pursuant to the terms of the relevant commercial agreement.
...
+EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO ANY WARRANTIES OF
+MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT
+OF COPYRIGHT, PATENT, TRADEMARK, OR OTHER RIGHT. IN NO EVENT SHALL THE
+COPYRIGHT HOLDER BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+INCLUDING ANY GENERAL, SPECIAL, INDIRECT, INCIDENTAL, OR CONSEQUENTIAL
+DAMAGES, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+FROM, OUT OF THE USE OR INABILITY TO USE THE FONT SOFTWARE OR FROM
+OTHER DEALINGS IN THE FONT SOFTWARE.
+
+=========================================================================
```
jodok: delimiter

LICENSE
```diff
@@ -217,3 +200,177 @@ solely pursuant to the terms of the relevant commercial agreement.
...
+
+For the `bootstrap.py` file:
```
jodok: this refers not only to the bootstrap.py - it's the package "Buildout"

NOTICE
```diff
@@ -1,98 +1,13 @@
...
+under one or more contributor license agreements.  See the NOTICE file
```
jodok: we don't need the reference to the NOTICE file as this has been moved here.

bootstrap.py
```diff
@@ -35,7 +35,7 @@
 Simply run this script in a directory containing a buildout.cfg, using the
 Python that you want bin/buildout to use.
 
-Note that by using --find-links to point to local resources, you can keep 
+Note that by using --find-links to point to local resources, you can keep
```
jodok: can you revert all changes to the file?
it's a boilerplate file that shouldn't be modified at all (despite the trailing spaces i dislike as well :))

PR #555, opened 2018-04-06
various smaller cleanups

chaudum:


app/views/tables.html
```diff
@@ -1,9 +1,13 @@
 <div>
   <div ng-class="$root.showSideList ? '' : 'hide-sidebar-list'">
   	<div class="page-container" ng-if="empty">
-      <h1 class="cr-page-header">Tables</h1>
+      <h1 class="cr-page-header">{{ 'TABLE.TITLE' | translate }}</h1>
       <div class="cr-panel-block">
-        There are no tables created yet.
+        <p>
+          <b>There are no tables created yet.</b>
```
Autophagy: Would prefer to wrap these in their on `<p></p>` tags rather than using `<br /><br />`, but its just a personal preference. I think the text in the Help tab does it this way.

app/views/views.html
```diff
@@ -1,9 +1,13 @@
...
-  <div ng-class="$root.showViewList ? '' : 'hide-sidebar-list'">
+  <div ng-class="$root.showSideList ? '' : 'hide-sidebar-list'">
     <div class="page-container" ng-if="empty">
       <h1 class="cr-page-header">{{ 'VIEW.TITLE' | translate }}</h1>
       <div class="cr-panel-block">
-        There are no views created yet.
+        <p>
+          <b>There are no views created yet.</b>
+          <br /><br />
+          Use the <a href="https://crate.io/docs/crate/reference/en/latest/general/ddl/create-view.html">CREATE VIEW</a> statement to create new views.
```
chaudum: Change URL to `.../views.html` see https://github.com/crate/crate/pull/7138/files#diff-473eca00626203e4bf5df7059061162eR26

PR #556, opened 2018-04-09
fill out the docs

nslater:


docs/cluster.txt
```diff
@@ -1,5 +1,158 @@
...
+
+  The number of started (i.e. available) shards.
+
+:Reallocating:
+
+  The number of shards that are being moved to another node.
+
+:Post Recovery:
+
+  FIXME: what is this?
```
nslater: @mxm can you help?

docs/cluster.txt
```diff
@@ -1,5 +1,158 @@
...
+
+:Reads:
+
+  FIXME: what is this?
+
+:Writes:
+
+  FIXME: what is this?
```
nslater: @mxm here

docs/cluster.txt
```diff
@@ -1,5 +1,158 @@
...
+:Post Recovery:
+
+  FIXME: what is this?
+
+Disk Operations
+---------------
+
+:Read:
+
+  FIXME: what is this?
```
nslater: @mxm here

docs/cluster.txt
```diff
@@ -1,5 +1,158 @@
...
+
+:Write:
+
+  FIXME: what is this?
```
nslater: @mxm here

docs/cluster.txt
```diff
@@ -1,5 +1,158 @@
...
+
+:Reads:
+
+  FIXME: what is this?
```
nslater: @mxm here

docs/shards.txt
```diff
@@ -1,5 +1,101 @@
...
+
+  :Started |nbsp| Replica: (dark green)
+
+     This is a `replica`_ shard and has been successfully started.
+
+  :Initializing: (dark yellow)
+
+     The shard is being initialized (i.e. started).
+
+     FIXME: is this true?
```
nslater: @mxm here

docs/shards.txt
```diff
@@ -1,5 +1,101 @@
...
+  :Relocating: (dark red)
+
+     The shard is being relocated to another node.
+
+     CrateDB automatically rebalances your cluster for you. While a shard is
+     being moved to another node, it is unavailable for use.
+
+  :Unassigned: (gray)
+
+     FIXME: what does this mean?
```
nslater: @mxm here

docs/tables.txt
```diff
@@ -1,5 +1,147 @@
...
+
+:Name:
+
+  The name of the table.
+
+:Health:
+
+  The health of the table.
+
+  FIXME: what are the available options
```
nslater: @mxm here

docs/tables.txt
```diff
@@ -1,5 +1,147 @@
...
+
+  The number of configured `replicas`_.
+
+:Configured Shards:
+
+  The `number of configured shards`_
+
+:Started Shards:
+
+  FIXME: what is this?
```
nslater: @mxm here

docs/tables.txt
```diff
@@ -1,5 +1,147 @@
...
+
+:Missing Shards:
+
+  FIXME: what is this?
```
nslater: @mxm here

docs/tables.txt
```diff
@@ -1,5 +1,147 @@
...
+
+:Underrepl. Shards:
+
+  FIXME: what is this?
```
nslater: @mxm here

docs/tables.txt
```diff
@@ -1,5 +1,147 @@
...
+
+:Total Records:
+  The total number of records (i.e. rows) the table has.
+
+:Unavailable Records:
+
+  FIXME: what is this?
```
nslater: @mxm here

docs/tables.txt
```diff
@@ -1,5 +1,147 @@
...
+
+:Underrepl. Records:
+
+  FIXME: what is this?
```
nslater: @mxm here

docs/tables.txt
```diff
@@ -1,5 +1,147 @@
...
+
+:Size:
+
+  The total size on disk used by all primary shards.
+
+:Recovery:
+
+  FIXME: what is this?
```
nslater: @mxm here

nslater:
@mxm I've left a bunch of comments on my changeset. not sure this is the best way to request your input on my changes, but let's see!!

basically, every FIXME just needs some input because I am not certain myself what needs to go there / what the correct description/explanation is

docs/index.txt
```diff
@@ -100,20 +101,18 @@ Running down the left-hand side, from top to bottom, the tabs are:
 
 - `Overview screen`_
 
-- :ref:`SQL console <console>`
-
-- :ref:`Tables browser <tables>`
+- :ref:`SQL console <sql-console>`
 
-- :ref:`Views browser <views>`
```
chaudum: The views tab has been added with the last big commit.

docs/cluster.txt
```diff
@@ -1,5 +1,158 @@
...
+
+:Disk Usage:
+
+  A visual indicator of system disk space utilization.
+
+:CrateDB CPU Usage:
+
+  A visual indicator of allocated CrateDB CPU utilization.
+
+  TODO: is this correct? needs more explaining I think
```
nslater: @mxm oh, here too

docs/index.txt
nslater: ah damn! ok. I'll re-download. ty

chaudum: :-)

docs/cluster.txt
mxm: Shards which data have been successfully recovered from disk (e.g. after starting a new node).

mxm: This is from the `sys.shards` table.

docs/cluster.txt
mxm: This is also available via `sys.nodes`.

docs/cluster.txt
mxm: Total number of read sizes. This is deprecated. See https://crate.io/docs/crate/reference/en/latest/admin/system-information.html#fs

mxm: This is also available via `sys.nodes`.

docs/cluster.txt
mxm: The number of read operations.

docs/cluster.txt
mxm: The number of write operations.

docs/shards.txt
mxm: it's starting but not started yet, e.g. recovering.

docs/shards.txt
mxm: A shard that has not been assigned to a node. Effectively, if the shard has any data, it won't be accessible in the cluster until it becomes started.

docs/tables.txt
mxm: This is red, yellow, or green.

>The cluster health status is: green, yellow or red. On the shard level, a red status indicates that the specific shard is not allocated in the cluster, yellow means that the primary shard is allocated but replicas are not, and green means that all shards are allocated. The index level status is controlled by the worst shard status. The cluster status is controlled by the worst index status.

https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-health.html

docs/tables.txt
mxm: Started means the shard is assigned to a node and available for querying.

docs/tables.txt
mxm: Shards which data has been deleted. The shard is still registered in the meta data.

docs/tables.txt
mxm: Shards which don't have the configured number of replicas.

docs/tables.txt
mxm: The number of records which were saved in the unavailable shards.

docs/tables.txt
mxm: The number of records which are currently don't have a replica copy.

docs/tables.txt
mxm: The percentage of the recovery process completed (e.g. when starting the cluster).

docs/index.txt
nslater: this is too much work for now, I'll make a note and add this page once the changes land on crate/crate master

docs/cluster.txt
nslater: I'm not sure I follow

if I stop a node and then start it again, are all the shards on that node marked as "post-recovery"? (and if so, for what duration of time?)

or is there some subtlety I'm missing?

docs/cluster.txt
nslater: total or per time period?

if total, is it total since CrateDB was last started?

docs/cluster.txt
nslater: I have added a CAUTION admonition saying:

> These values are read from the sys.nodes fs object, which is now deprecated. For the time being, this section is non-function.

is this correct?

nslater:
@mxm GitHub has hidden my comment, but I have left one here:

https://github.com/crate/crate-admin/pull/556#discussion_r180087993

docs/shards.txt
nslater: oh boy this is confusing. we have three terms for the same thing:

- starting
- recovering
- initializing

am I correct? eep

docs/tables.txt
nslater: why might this happen?

docs/cluster.txt
mxm: The `fs` object is not deprecated, only the `read/reads` and `write/writes` values.

docs/cluster.txt
nslater: hmm. are you sure? a few items above this we have "CPU Usage" which I have defined as system CPU usage. from the doc you've linked, wouldn't this be `getProcessCpuLoad`?

docs/cluster.txt
nslater: hmm. do you know why all four values show up as "-"?

docs/shards.txt
mxm: Yes, they are two separate concepts. The shard's physical data can be recovered but it might not yet be accessible via cluster routing.

docs/tables.txt
```diff
@@ -1,5 +1,171 @@
...
+
+  :Yellow:
+
+    The primary shard is allocated, but not all replica shards are allocated.
+
+  :Red:
+
+    This specific shard is not allocated.
+
+  FIXME: add brief explanation of what allocated means
```
nslater: @mxm can you help with this? I mentioned this on another comment

docs/tables.txt
```diff
@@ -1,5 +1,171 @@
...
+
+  The `number of configured shards`_
+
+:Started Shards:
+
+  The number of started shards assigned to the node and available for querying.
+
+:Missing Shards:
+
+  FIXME: what is this?
```
nslater: hmm. the comment on this has vanished. so putting one back haha

nslater:
@mxm okay pushed some fixups and left some comments! ty ty ty!

docs/shards.txt
mxm: Started means it is available in the cluster. Initializing is actually a CrateDB term which means a shard has been recovered but has not been assigned yet.

docs/tables.txt
mxm: You delete your data directory on a node.

docs/tables.txt
mxm: Allocated here means that it is fully available to the cluster.

docs/tables.txt
mxm: Shards for which the shard data is missing. (e.g. you delete your data directory) and they could not be recovered.

docs/cluster.txt
mxm: More context:
https://www.elastic.co/guide/en/elasticsearch/reference/1.4/cat-recovery.html

>A recovery event occurs anytime an index shard moves to a different node in the cluster. This can happen during a snapshot recovery, a change in replication level, node failure, or on node startup. This last type is called a local gateway recovery and is the normal way for shards to be loaded from disk when a node starts up.

nslater: thank you!

docs/tables.txt
nslater: by unavailable here do you mean not allocated or missing?

nslater:
@mxm I have pushed some more changes and squashed my fixups. could you take a look through https://github.com/crate/crate-admin/pull/556/files and check it for technical accuracy re my definitions of things, and also, any outstanding comment threads. thanks so much!

sorry this one is such a hassle!

I think there's some work to be done in the main docs defining these terms. as someone who's not familiar with ES, our docs assume a lot of prior knowledge/vocab

docs/cluster.txt
mxm: `POST_RECOVERY` is the state of the shard when it has been recovered but is not yet `STARTED`. `STARTED` means it is accessible in the cluster through the cluster routing. It is the desired state of the shard because only then you can read/write data to it.

docs/tables.txt
mxm: AFAIK the records of the missing shards.

docs/cluster.txt
mxm: Sorry, I was looking at the CPU usage above this. You're correct, this should be the process cpu usage.

PR #562, opened 2018-04-18
updated wording and removed abbreviations

Taliik:
removed the abbreviations in the overview & table view
in nodes view the detail view "nodes" is changed to "overview" since it only shows details of a single node. Renamed the label for the REST URL to HTTP Endpoint


m-kharbat:
I think a changes entry is missing.. please also backport to 1.8 and 1.9

PR #563, opened 2018-04-19
Fixed several issues concerning table/view filtering

### LINKED ISSUE #560: Clicked Table doesnt show the Table
Hi there, 
it seems like the Filtering of Tables described in #535 seems to work now in Crate 2.3.6 but I have noticed that the link to the tables are still completely wrong. 

As You can see here if I search for "host" Tables and click on statusengine_hoststatus I get the "openitcockpit_contacts" Table. 

Filtered List
broken

When I remove the Filter (the reset filter button in the input field is also not working) you can see the original order of all tables. Seems to me like the links of the filtered list is still in the order as the unfiltered list so everything that is on position 1 after filtering get the link to the table at position 1 in the unfiltered list. 

Unfiltered List
bildschirmfoto 2018-04-17 um 11 30 09

Filtered List
broken2

Browser: Firefox 59.0.2 
Crate: 2.3.6

Autophagy:
When filtering the table/view list, the incorrect link, table status and table size would be preserved from what was previously in that position in the table/view list. This fix changes this so that the links/status/size are updated as we would expect.

The clear filter button also was not working. This has also been fixed.

This should fix #560

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,6 +5,10 @@ Changes for Crate Admin Interface
 Unreleased
 ==========
 
+ - Fixed several bugs concerning filtering tables/views where table
+   information would not be updated properly. Also fixed the clear filter
+   button for both table and view filtering.
+
  - Removed abbreviations in the overview and table view. Updated nodes view.
 
  - Remove trailing space from column name in tables detail view.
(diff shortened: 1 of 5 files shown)


Autophagy:
@m-kharbat This kind of undos some of your work in reducing the amount of watchers https://github.com/crate/crate-admin/commit/f539a205f8329d2828432f8004503b4563c394ce - but I'm not sure how we would update these dynamically without them.

PR #572, opened 2018-05-16
Meri/fix errors 3.0

m-kharbat:


m-kharbat:
no, they are not user facing issues

PR #575, opened 2018-05-25
Fixed an issue that caused the CrateDB process CPU calculation to be …

m-kharbat:
…false

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,6 +5,8 @@ Changes for Crate Admin Interface
 Unreleased
 ==========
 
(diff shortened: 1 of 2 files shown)


CHANGES.txt
```diff
@@ -5,6 +5,8 @@ Changes for Crate Admin Interface
 Unreleased
 ==========
 
+- Fixed an issue that caused the CrateDB process CPU calculation to be false.
```
chaudum: Think the correct term is "wrong" (not "false").

PR #577, opened 2018-05-30
improve use of admonitions

nslater:
okay let's try again on a different project 😬

I just added a new set of styles for admonitions in the docs. I am now going through each docs project and improving our use of admonitions

### DIFF
--- a/docs/help.txt
+++ b/docs/help.txt
@@ -33,6 +33,6 @@ Features
   When you are satisfied with the number of imported tweets, select the
   *ENOUGH! STOP IMPORTING TWEETS* button.
 
-  .. NOTE::
+  .. TIP::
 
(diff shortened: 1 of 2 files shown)


docs/help.txt
```diff
@@ -33,6 +33,6 @@ Features
   When you are satisfied with the number of imported tweets, select the
   *ENOUGH! STOP IMPORTING TWEETS* button.
 
-  .. NOTE::
+  .. TIP::
```
emilywoods: Why was this one changed to `TIP`? To me, `NOTE` seems rather appropriate for it :)

nslater: hmm! I think it's because I am using NOTE for things you should read and TIP for things that you don't have to read but you might be interested in. if that makes sense? :)

PR #580, opened 2018-07-04
Meri/upgrade libs

m-kharbat:
- upgraded dependency.
- migrate to webpack 4 to reduce vulnerabilities


app/scripts/controllers/common.js
```diff
@@ -153,8 +153,10 @@ const commons = angular.module('common', ['stats', 'udc', 'events', 'sql'])
     updateClusterInfo();
 
     // bind tooltips
-    $('[rel=tooltip]').tooltip({
-      placement: 'bottom'
+    $(document).ready(function() {
```
autophagy: Why is waiting for the document ready needed now? Or was it always causing issues?

m-kharbat: i forgot to remove this, I will push a fix up.
the waiting for document ready is not needed when jquery and bootstrap are loaded properly

autophagy: Ah ok! Cool

PR #583, opened 2018-08-22
fix(navigation): Fixed the navigation component to allow opening URLs…

m-kharbat:
… in new tabs

using <a> tag allows the browser to understand that you can open the link in a new tab

screen shot 2018-08-22 at 15 45 43


autophagy:
@m-kharbat Take a look at those failing tests first, though, please :innocent:

PR #589, opened 2018-11-05
Change license information

smakalias:
Changed the license information (ident) to be taken from the `sys.cluster.licence`
 attribute instead of the `license.ident` setting which is `@deprecated`.


app/app.module.js
```diff
@@ -148,8 +148,11 @@ $.get('static/conf/plugins.json', function (plugins) {
         $stateProvider, SettingsProvider) {
 
         var stmt = 'SELECT settings[\'license\'][\'enterprise\'] as enterprise, ' +
-          'settings[\'license\'][\'ident\'] as ident ' +
-          'FROM sys.cluster';
+//                    'settings[\'license\'][\'ident\'] as ident ' +
+                        'if(license is not NULL, ' +
+                            'format(\'%s (exp: %tD)\', license[\'issuedTo\'], license[\'expirationDateInMs\']), ' +
```
mfussenegger: please use `%tF` for ISO date instead of US format that nobody else uses.

app/app.module.js
```diff
@@ -148,8 +148,11 @@ $.get('static/conf/plugins.json', function (plugins) {
         $stateProvider, SettingsProvider) {
 
         var stmt = 'SELECT settings[\'license\'][\'enterprise\'] as enterprise, ' +
-          'settings[\'license\'][\'ident\'] as ident ' +
-          'FROM sys.cluster';
+//                    'settings[\'license\'][\'ident\'] as ident ' +
```
mfussenegger: remove?

app/app.module.js
smakalias: done

app/app.module.js
smakalias: done

PR #620, opened 2019-04-03
Adapt query to schema_name → table_schema change

mfussenegger:


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "crate-admin",
   "description": "Crate Admin UI",
-  "version": "1.15.0-dev",
+  "version": "1.14.1-dev",
```
mfussenegger: Is 1.14 already for CrateDB specific or do we have to go with 1.15 ?

seut: It is already used in 3.3.

PR #630, opened 2019-07-10
Rename schema_name to tables_schema in tables partion query

mkleen:
information_schema.table_partitions.schema_name was renamed
to information_schema.table_partitions.table_schema in CrateDB
4.0.0

## Checklist

 - [X] [CLA](https://crate.io/community/contribute/cla/) is signed

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,16 @@ Changes for Crate Admin Interface
 Unreleased
 ==========
 
+
+2019/07/10 1.15.1
+=================
+
+Fixes
+-----
+
+- Adapted a query for compatibility with the upcoming CrateDB 4.0 release.
+
+
 2019/04/03 1.15.0
 =================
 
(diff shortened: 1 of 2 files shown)


app/scripts/controllers/tables.js
```diff
@@ -108,7 +108,7 @@ angular.module('tables', ['stats', 'sql', 'common', 'tableinfo', 'events'])
 
         const PARTITION_QUERY = 'SELECT partition_ident, number_of_shards, number_of_replicas, values ' +
           'FROM information_schema.table_partitions ' +
-          'WHERE schema_name = ? AND table_name = ? AND closed = false';
+          'WHERE table_schema = ? AND table_name = ? AND closed = false';
```
m-kharbat: is this a breaking change ? it's missing a Changes.txt entry

mkleen: Thanks, it is, I will add it.

CHANGES.rst
```diff
@@ -5,6 +5,11 @@ Changes for Crate Admin Interface
 Unreleased
 ==========
 
+Breaking Changes
```
mfussenegger: I think this could go as fix within `1.15`, given that `1.15.0` already broke compatibility and is for 4.0+

mkleen: ok, that would be then version 1.15.1 ?

Issue #634, opened 2019-08-22
Bad columns view for tables on 4.0.4 cluster


shm-dmitry:
**Enviroment**: crate 4.0.4 with default configuration (3.3.5 works fine)

**Problem description**:
All columns showed as _generated_
See screenshot in attach (red square)

![Screenshot_20190822_133527](https://user-images.githubusercontent.com/19342331/63508185-1b3ac480-c4e2-11e9-92b3-ee7abe6887b8.png)

**Steps to reproduce**:
`create table test (a string, b int);`
`>>  CREATE TABLE OK, 1 row affected (0.346 sec)`

`show create table test;`
```
CREATE TABLE IF NOT EXISTS "doc"."test" (
   "a" TEXT,
   "b" INTEGER
)
CLUSTERED INTO 4 SHARDS
WITH (
   "allocation.max_retries" = 5,
   "blocks.metadata" = false,
   "blocks.read" = false,
   "blocks.read_only" = false,
   "blocks.read_only_allow_delete" = false,
   "blocks.write" = false,
   column_policy = 'strict',
   "mapping.total_fields.limit" = 1000,
   max_ngram_diff = 1,
   max_shingle_diff = 3,
   number_of_replicas = '0-1',
   refresh_interval = 1000,
   "routing.allocation.enable" = 'all',
   "routing.allocation.total_shards_per_node" = -1,
   "translog.durability" = 'REQUEST',
   "translog.flush_threshold_size" = 536870912,
   "translog.sync_interval" = 5000,
   "unassigned.node_left.delayed_timeout" = 60000,
   "warmer.enabled" = true,
   "write.wait_for_active_shards" = 'ALL'
)
```

PR #635, opened 2019-08-27
Adapt column query for `is_generated` change in CrateDB 4.0

mfussenegger:
is_generated changed in 4.0 from a boolean to a string (`ALWAYS`,
`NEVER`) to conform to the SQL standard.

Fixes https://github.com/crate/crate-admin/issues/634

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,9 @@ Changes for Crate Admin Interface
 Unreleased
 ==========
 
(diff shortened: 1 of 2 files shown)


mfussenegger:
> please update the formatting

Which formatting and how?

I intentially didn't indent the query so that the statement won't have a lot of extra spaces in it and won't look weird on CrateDB side in the sys.jobs_log.

m-kharbat:
okay, makes sense

PR #637, opened 2019-09-05
Export plugin

rps-v:
Added **export** plugin with a UI similar to Console.
From here you can now export queries in the following formats: JSON, CSV, CSV (flattened - object properties become own columns).

**NOTE**: This plugin was made for version 1.14.0.

image

## Checklist


mfussenegger:
Thank you a lot for the contribution.

What do you think about integrating this export functionality directly into the regular console view? Maybe having the inline table display as the default output options, and json/csv, etc. as alternatives.

rps-v:
I actually thought about that, but didn't want to mess with the built-in plugins. But now that you mention it, I'll give it a try.

rps-v:
Moved functionality to Console.

image

app/scripts/controllers/console.js
```diff
@@ -45,7 +45,78 @@ const crate_console = angular.module('console', ['sql', 'datatypechecks', 'stats
 
     return service;
   })
-  .directive('console', function(SQLQuery, ColumnTypeCheck, ConsoleFormatting, ClusterState){
+  .service('ExportUtils', function () {
+    var utils = {};
+
+    utils.arrayFrom = function (json) {
```
m-kharbat: It would be great if you add tests to the functions of this service. We keep the test files in /tests folder

app/scripts/controllers/console.js
```diff
@@ -128,6 +205,67 @@ const crate_console = angular.module('console', ['sql', 'datatypechecks', 'stats
...
+    
+                row.forEach((column, index) => {
+                  var last = index === row.length - 1;
+                  var type = getDataType(column, col_types[index]);
+    
+                  switch (type) {
+                    case 'object':
+                    case 'array':
+                      // Stringify twice to escape the output
+                      output += JSON.stringify((JSON.stringify(column)));
```
m-kharbat: I don't understand why stringify is applied twice ? 🤔 Maybe you can add a test for this case ?

m-kharbat: okay I see, then It would be great if you test the export functionality for different cases to see that it matches the expected csv or json output

m-kharbat:
Thank you for contributing and updating the pr @rps-v 🎉 
regarding the tests let me know if you prefer that our team takes over, we can then merge you pr to a feature branch.

by the way have you already signed the contributor CLA agreement ? 🍸 https://crate.io/community/contribute/cla/

rps-v:
Yes, I think it would be best if your team takes over. And yes, I signed the CLA.

PR #641, opened 2019-09-24
Forces the query table results in the Console view to rerender when …

m-kharbat:
clicking on previous/next buttons

uses safeApply to force the $digest cycle and re-render the page

Object and Arrays are not expanded by default.
When the panigation next/previous are clicked, all objects and arrays are closed.
This forces the UI to not reuse DOM elements, when the user expands an object or array the ng-if ensure that correct data is rendered. 

## Checklist

 - [ ] [CLA](https://crate.io/community/contribute/cla/) is signed

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,9 @@ Changes for Crate Admin Interface
 Unreleased
 ==========
 
+- Fixed an issue that caused the result table to display incorrect results in
+  certain columns after clicking the pagination buttons.
+
 - Fixed an issue that caused the Twitter tutorial to not start automatically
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -5,6 +5,9 @@ Changes for Crate Admin Interface
 Unreleased
 ==========
 
+- Forces the query table results in the Console view to rerender when clicking
```
chaudum:
```suggestion
- Fixed an issue that caused the result table to display incorrect results in certain columns after clicking the pagination buttons.
```

PR #658, opened 2020-01-16
Update keywords in AdminUI

marregui:
So that cratedb's latest keywords are recognised and painted in red.


DEVELOP.rst
```diff
@@ -19,6 +19,7 @@ Installing node locally::
 
 Install the package dependencies::
 
+    $ export PATH="$PATH:/<absolute-path>/crate-admin/env/bin"
```
mfussenegger: The changes here shouldn't be necessary. Maybe what is missing is to activate the virtualenv properly. Though since the calls are made directly via `env/bin/...` that also shouldn't be necessary.

marregui: reverted

CHANGES.rst
```diff
@@ -19,6 +19,10 @@ Fixes
 - Fixed an issue that caused the node health to not be displayed in the Cluster 
   View when the node name was too large.
 
+- Updated keywords list so that they are recognised and painted in red.
```
mfussenegger: Why is this added beneath a already released version?

marregui: keen eye!

autophagy:
Should we also add a changelog entry for the `node-sass` version bump? Was this done to fix a vuln @marregui?

marregui:
Thank you!, it is the first time I work on a node project, thought the package.json file was some artifice of running the code, or something required by the ide...

CHANGES.rst
```diff
@@ -7,6 +7,11 @@ Fixes
 
 - Fixed capitalization of ``Shards`` tab label
 
+- Updated keywords list so that they are recognised and painted in red.
+
+- Upgraded ``node-sass``.
```
mfussenegger: If we add such dependency upgrades to the `CHANGES` we usually also include a reason. What is the effect of the upgrade for an end-user?

Issue #657, opened 2020-01-13
Wrong display of column value for nested partition key column


shm-dmitry:
Crate version: 4.0.10

*Steps to reproduce*

```
CREATE TABLE test
(
    pk object as (
        id string,
        part string
    ),
    PRIMARY KEY (pk['id'], pk['part'])
)
PARTITIONED BY (pk['part']);
>> (success)

insert into test (pk) values ( { id='1', part='x' } ) ;
>> (success)
```

Error marked on screenshots by red arrows:

1. In 'table view' does not showed partition key value:
![image1](https://user-images.githubusercontent.com/19342331/72256400-2bcb1300-361a-11ea-8a3e-247479aaee32.png)

2. In 'Query view' does not showed partition key value when column viewed as nested column.
But when I query column value directly - all data showed.
![image2](https://user-images.githubusercontent.com/19342331/72256411-31285d80-361a-11ea-8d9e-8010954b2855.png)

mfussenegger:
Thank you for the report. We'll have a closer look at it.

PR #660, opened 2020-01-29
Fix value display for nested partitioned columns

mfussenegger:
The partitioned by columns are fetched from `information_schema.tables`
where the column names use a dot notation:

    select partitioned_by from information_schema.tables where table_name = 'test';
    +----------------+
    | partitioned_by |
    +----------------+
    | ["pk.part"]    |
    +----------------+

These column names were then used to lookup the value from the `values`
in `information_schema.table_partitions`, but there the columns have the
subscript notation:

    select values from information_schema.table_partitions where table_name = 'test';
    +---------------------+
    | values              |
    +---------------------+
    | {"pk['part']": "x"} |
    +---------------------+

Fixes https://github.com/crate/crate-admin/issues/657

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,9 @@ Changes for Crate Admin Interface
 Fixes
 -----
 
+- Fixed an issue that prevents the value for nested partitioned columns showing
+  up in the table partitions overview.
+
 - Fixed capitalization of ``Shards`` tab label
 
 - Updated keywords list so that they are recognised and painted in red.
--- a/app/scripts/services/tableinfo.js
+++ b/app/scripts/services/tableinfo.js
@@ -7,9 +7,16 @@ const tableinfo = angular.module('tableinfo', ['sql'])
     var isActiveShard = function(shard) {
       return ACTIVE_SHARDS.indexOf(shard.routing_state) > -1;
     };
+
+    var dotNotationToSubscript = function (column) {
+      return column.replace(/\.([a-zA-Z]+)/g, function(match, p1, offset, str) {
+        return `['` + p1 + `']`; 
+      });
+    };
+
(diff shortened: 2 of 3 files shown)


CHANGES.rst
```diff
@@ -5,6 +5,9 @@ Changes for Crate Admin Interface
 Fixes
 -----
 
+- Fixed an issue that prevented the value for nested partitioned columns to
+  show up in the table partitions overview.
```
mechanomi: "columns showing up in the..." would be more correct. the use of "prevented" earlier in the sentence means that this part of the sentence should also use the past tense

alternatively, you could change the tense to present indicative (which is probably the better option):

"Fixed an issue that prevents the value for nested partitioned columns showing up in the table partitions overview."

app/scripts/services/tableinfo.js
```diff
@@ -7,9 +7,16 @@ const tableinfo = angular.module('tableinfo', ['sql'])
     var isActiveShard = function(shard) {
       return ACTIVE_SHARDS.indexOf(shard.routing_state) > -1;
     };
+
+    var dotNotationToSubscript = function (column) {
+      return column.replace(/\.([a-zA-Z]+)/, function(match, p1, offset, str) {
+        return `['` + p1 + `']`; 
+      });
```
seut: what about nested subscript expression like `pk.a.b`? looks to as this still wouldn't work correctly

mfussenegger: Ah right, pushed a fixup:

```
"obj.a.b.c".replace(/\.([a-zA-Z]+)/g, function(match, p1, offset, str) {
        return `['` + p1 + `']`; 
      })

→

"obj['a']['b']['c']"
```

Issue #669, opened 2020-06-16
Bug: When clicking "Query Table" generated query syntax is wrong for object fields with whitespaces


proddata:
**CrateDB version**:
4.1.6

**Environment description**:
CrateDB Cloud (Dev-Instance - 3 nodes are / 3 CPUs / 3GB Ram

**Problem description**:

When clicking "Query Table" generated query syntax is wrong for object fields with whitespaces 
**Steps to reproduce**:

``` SQL
CREATE table nexo.test (obj OBJECT(DYNAMIC));
--CREATE TABLE OK, 1 row affected (0.852 sec)

INSERT INTO nexo.test VALUES('{ "field1" : 100 }');
-- INSERT OK, 1 row affected (0.076 sec)

INSERT INTO nexo.test VALUES('{ "field 2" : 100 }');
-- INSERT OK, 1 row affected (0.076 sec)

SELECT * FROM nexo.test limit 100;
-- SELECT OK, 2 rows in set (0.006 sec)
```

 In Admin UI go to table nexo.test and press 'Query Table'
``` SQL
-- the following query gets generated:
SELECT "obj","obj['field 2']" FROM "nexo"."test" LIMIT 100;
```

Which leads to an error, as the syntax is wrong ... should be
``` SQL
SELECT "obj","obj"['field 2'] FROM "nexo"."test" LIMIT 100;
```
Or even better, should be ..
``` SQL
SELECT "obj" FROM "nexo"."test" LIMIT 100;
```

proddata:
Regex in tables.js:144 is wrong / doesn't take care of whitespaces within columns
```javascript
var isNestedColumn = function (column) {
          var re = /([^\s]+)(\[\'([^\s]+)\'])+/i;
          return column.match(re);
        };
```

probably should be something like ...
```javascript
var isNestedColumn = function (column) {
          var re = /([^\s]+)(\[\'(.+)\'])+/i;
          return column.match(re);
        };
```

autophagy:
@proddata Thanks for reporting this and pinning down the issue! The fix has been merged, and will be available in the next admin ui release. :slightly_smiling_face:

PR #671, opened 2020-06-22
Fixed issue in nested-column identification regex

autophagy:
This should address #669

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,16 @@
 Changes for Crate Admin Interface
 =================================
 
+Unreleased
+==========
+
+Fixes
+-----
+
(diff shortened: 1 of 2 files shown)


CHANGES.rst
```diff
@@ -2,6 +2,15 @@
 Changes for Crate Admin Interface
 =================================
 
+Unreleased
+==========
+
+Fixes
+-----
+
+- Fixed an issue where a nested column with a space in it would not be as one,
```
mfussenegger, line 11: The "not be as one" part may be a bit confusing without more context.

autophagy: Aha, im missing the word "identified" :see_no_evil:

Issue #667, opened 2020-06-08
CrateDB has new type `time with time zone`


marregui:
With this PR https://github.com/crate/crate/pull/9915 

the type travels as an array containing two values:

- microseconds from midnight
- signed seconds from UTC

E.g. [53388674000, 0]

The Admin UI should be able to render it.

PR #673, opened 2020-06-23
Add results formatting for new Time with Timezone type

autophagy:
This addresses #667

### DIFF
--- a/app/scripts/filter/text.js
+++ b/app/scripts/filter/text.js
@@ -77,9 +77,42 @@ const filters_text = angular.module('filters_text', [])
     return function(timestamp) {
       var is_valid = timestamp == 0 || (new Date(timestamp)).getTime() > 0;
       if (!is_valid) {
-        return "Invalid Timestamp";
+        return 'Invalid Timestamp';
       }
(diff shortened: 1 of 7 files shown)


autophagy:
@marregui Requested your review on this, in case you had a nicer idea about how to format microseconds from midnight and seconds from UTC into a string. I did it this way because I'd rather not add an extra time library if I can help it. :sweat_smile:

app/scripts/filter/text.js
```diff
@@ -77,10 +77,43 @@ const filters_text = angular.module('filters_text', [])
...
       }
       return new Date(timestamp).toISOString();
     };
+  })
+  .filter('formatTimeWithTimezone', function() {
+    function formatTimeElement(element) {
+      if (element < 10) {
+        return '0' + Math.floor(element);
+      } else {
+        return Math.floor(element);
```
mfussenegger, line 90: Would it make sense to force a string conversion in the second branch too, to avoid having different types returned?

autophagy: Yeah, it'd be good for consistencies sake :+1:

PR #684, opened 2020-08-20
Tweet tutorial table no longer uses deprecated column types

autophagy:
This, previously, would cause deprecated error messages on the CrateDB node due to the use of the `TIMESTAMP` type.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,14 @@
 Changes for Crate Admin Interface
 =================================
 
+Unreleased
+==========
+
+Changes
+-------
+
+- Updated the Tweet tutorial table to no longer use deprecated column types.
+
 2020-07-01 1.16.0
 =================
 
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -2,6 +2,14 @@
 Changes for Crate Admin Interface
 =================================
 
+Unreleased
+==========
+
+Fixes
+-----
+
+- Updated the Tweet tutorial table to no longer use deprecated column types.
```
kovrus, line 11: I am wondering whether it is a fix.

kovrus (approved):

mechanomi (commented):
lgtm

CHANGES.rst
autophagy: Do you think it counts more as a 'change'?

autophagy:
@mechanomi Any insight as to why the travis build might be failing here?

PR #690, opened 2020-09-18
Remove tweet tutorial demo.

seut:
It is currently broken and also does not reflect the most common use cases with CrateDB.
We may come up with a better demonstration in future.


mechanomi (changes requested):
the Help Screen documentation should be stripped down to match

don't bother with replacing the screenshot, as I already have an outstanding PR for updating those. but it would be good to update the text on that page

see: https://crate.io/docs/crate/admin-ui/en/latest/help.html#help-screen

Perhaps something simple like:

> This page includes links to various help resources.

(under the "Features" heading)

it's not much. but it's better than nothing. I would be a -0 on including any text promising a new feature here "in a future" release. but if you want to do it, I wouldn't be -1 on it

seut:
@mechanomi Good catch, thx for pointing out.
I've updated the help documentation text and also the screenshot so all is aligned now (having a still wrong screenshot seems awkward to me). Of course the updated screenshot may requires some polishing by you ;-)

mechanomi (approved):

Issue #686, opened 2020-08-23
"HTTP endpoint" in the UI should be written in title case


mechanomi:
how to reproduce:

- visit the cluster browser in the Admin UI
- select a node

what I see:

- a bunch of node info fields all using title case except for "HTTP endpoint"

what I expect to see:

- "HTTP Endpoint" to match the style of the other fields

the docs need updating to match when this change is made

PR #692, opened 2020-09-24
Correct uppercasing of 'HTTP Endpoint'

autophagy:

This should address #686 

@mechanomi When you say "the docs need updating to match when this change is made" - what/where exactly do you mean?


mechanomi:
@autophagy I'm referring to the screenshots! does that clear it up?

mechanomi (approved):
looks good. but as previously commented: screenshots should be updated too

autophagy:
Okidoki!

PR #696, opened 2020-10-12
Correct version number in changes

mfussenegger:
🤦


seut (approved):
I also didn't recognize 🤦

PR #699, opened 2020-11-10
Replace Travis CI docs tests with GitHub Actions workflow

mechanomi:
replace Travis CI (docs tests only) with GitHub Actions workflow - based on https://github.com/crate/crash/pull/341#issuecomment-700773923

This change reverts `.travis.yml` to the version prior to the addition of the
docs tests (i.e., bdbaa967d31c43d3ec35bd9457c6b5cbd1c30a3c).

todo list:

- [ ] merge this PR
- [ ] enable GitHub Actions workflow branch protection

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,44 +1,14 @@
 notifications:
   email: false
-
-code: &code
-  stage: "Code"
-  language: node_js
-  before_install:
-    - pip install --user codecov
-  before_script:
-    - "npm install"
-  script:
-    - npm run travis
-  branches:
-    except:
-      - /.*\/.*/
-
-docs: &docs
-  stage: "Documentation"
-  language: generic
-  script:
-    - cd docs && make check
-
-matrix:
-  include:
-
-    - <<: *code
-      name: "Node 8.9.0"
-      node_js:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,44 +1,12 @@
-notifications:
-  email: false
```
autophagy, line 2: Is this false by default, now?

autophagy (approved):
We could also take a look at completely removing the travis from this repo and running the codecov stuff on GH actions, but lgtm!

.travis.yml
mechanomi: looks like by default it sends email (https://docs.travis-ci.com/user/notifications/#configuring-email-notifications) to the "committer and the author, but only if they have access to the repository the commit was pushed to"

my changeset reverts the file to how it was before I modified it to add docs testing. but I am happy to add this snippet back in if you think that is best (I am not familiar with what is typical here)

autophagy: I'd leave it in for now :slightly_smiling_face:

mechanomi:
pushed a fixup!

PR #702, opened 2021-01-28
Tutorial: Use "cratedb" tag when linking to Stack Overflow

amotl:
Hi there,

while I recognize CrateDB has been conceived before the advent of Rust's Cargo package manager and its ecosystem of Crates, we might want to adjust to reality that the tags on Stack Overflow associated with `crate` are more related to Rust these days and update this to use the tag `cratedb` instead. [1] shows that this space is already well populated.

While it might still be appropriate to use filesystem locations like `/var/lib/crate` to refer to directories where CrateDB is operated within local namespaces (https://github.com/crate/crate-howtos/pull/268), we should adjust this on public spaces like Stack Overflow in order to educate ourselves and our users.

With kind regards,
Andreas.

[1] https://stackoverflow.com/tags/cratedb

### DIFF
--- a/app/plugins/tutorial/tutorial.html
+++ b/app/plugins/tutorial/tutorial.html
@@ -10,7 +10,7 @@
     <div class="cr-panel-block--help">
       <h2>{{:: 'HELP.TITLE' | translate }} </h2>
       <div class="help-sources__wrapper">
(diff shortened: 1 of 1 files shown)


seut (approved):
good catch!

PR #707, opened 2021-02-16
Align with license change

amotl:
Hi there,

this aligns the code base with the upcoming license change. Enterprise features are now available to everyone.

With kind regards,
Andreas.


lukasbals (approved):
Looks good to me! 💯

infoverload (approved):

CHANGES.rst
```diff
@@ -8,6 +8,7 @@ Unreleased
 
 - Fixed translations. French now works again.
 - Fixed missing whitespace.
+- Align with license change. Enterprise features are now available to everyone.
```
seut, line 11: We normally put most recent changes on top. While on that could you also add empty lines between each entry please? thx!

app/app.module.js
```diff
@@ -125,77 +124,13 @@ var loadStylesheet = function (url) {
...
-      name: el.name,
-      files: [el.uri],
-      routing: el.routing,
-      stylesheet: el.stylesheet
-    };
-  });
-  DEFAULT_PLUGINS = plugins.filter(function (p) {
-    return !p.enterprise;
+  PLUGINS = plugins.filter(function (p) {
+    return p.enabled;
```
seut, line 128: As all plugins are enabled by default now, it looks like we could simply remove this feature and just list all plugins?
Afaik this whole filtered-plugin feature was only reasonable by enterprise plugin distinction..

app/plugins/monitoring/monitoring.css
```diff
@@ -1,21 +1,3 @@
...
- *
- * To use this file, Crate.io must have given you permission to enable and
- * use such Enterprise Features and you must have a valid Enterprise or
- * Subscription Agreement with Crate.io.  If you enable or use the Enterprise
- * Features, you represent and warrant that you have a valid Enterprise or
- * Subscription Agreement with Crate.io.  Your use of the Enterprise Features
- * if governed by the terms and conditions of your Enterprise or Subscription
- * Agreement with Crate.io.
- */
-
```
seut, line 18: I think we should add the Crate APL header here instead.

app/plugins/monitoring/monitoring.js
```diff
@@ -1,20 +1,3 @@
...
- * Unauthorized copying of this file, via any medium is strictly prohibited.
- *
- * To use this file, Crate.io must have given you permission to enable and
- * use such Enterprise Features and you must have a valid Enterprise or
- * Subscription Agreement with Crate.io.  If you enable or use the Enterprise
- * Features, you represent and warrant that you have a valid Enterprise or
- * Subscription Agreement with Crate.io.  Your use of the Enterprise Features
- * if governed by the terms and conditions of your Enterprise or Subscription
- * Agreement with Crate.io.
- */
```
seut, line 17: same here, use Crate APL header

app/plugins/monitoring/test/monitoringQueryService.test.js
```diff
@@ -1,20 +1,3 @@
...
- * Unauthorized copying of this file, via any medium is strictly prohibited.
- *
- * To use this file, Crate.io must have given you permission to enable and
- * use such Enterprise Features and you must have a valid Enterprise or
- * Subscription Agreement with Crate.io.  If you enable or use the Enterprise
- * Features, you represent and warrant that you have a valid Enterprise or
- * Subscription Agreement with Crate.io.  Your use of the Enterprise Features
- * if governed by the terms and conditions of your Enterprise or Subscription
- * Agreement with Crate.io.
- */
```
seut, line 17: and here (APL header) ;)

docs/index.rst
```diff
@@ -116,15 +111,14 @@ On the left-hand side, from top to bottom, the tabs are:
...
 - :ref:`Cluster browser <cluster-browser>`
-- :ref:`Monitoring overview <monitoring-overview>` (`enterprise feature`_)
-- :ref:`Privileges browser <privileges-browser>` (`enterprise feature`_)
+- :ref:`Monitoring overview <monitoring-overview>`
+- :ref:`Privileges browser <privileges-browser>`
 - :ref:`Help screen <help-screen>`
 
 
 .. _Cluster name: https://crate.io/docs/crate/reference/en/latest/config/node.html#basics
 .. _Community Edition: https://crate.io/docs/crate/reference/en/latest/editions.html#cratedb-community-edition
```
seut, line 122: is this still used?

seut (commented):
Looks good, added some suggestions.

app/app.module.js
amotl: I also tried to think about it and wanted to keep the filtering code for... reasons [1]. That's why I amended it to filter by and `enabled` configuration attribute. However, we can also remove this logic altogether.

[1] Maybe someone will come up with plugins we only want to enable for specific builds? However, I recognize this is completely hypothetical, just wanted to share my thoughts.

app/plugins/monitoring/monitoring.css
amotl: That also crossed my mind. I will just copy in the file header from the main CrateDB sources like [1], right?

[1] https://github.com/crate/crate/blob/master/server/src/main/java/io/crate/planner/consumer/InsertFromSubQueryPlanner.java

docs/index.rst
amotl: You mean the _monitoring_ and the _privileges_ views? Sure, both are two of the plugins which just have been available in the former enterprise build, see [1].

[1] https://github.com/crate/crate-admin/tree/master/app/plugins

app/app.module.js
seut: I'd remove it completely, we can re-add any time when there is a real need for.

app/plugins/monitoring/monitoring.css
amotl: > I think we should add the Crate APL header here instead.

All other files also lack respective headers. Shall we take care of that in a followup PR in order to keep the scope of this to semantic changes and the other patch for filling the gaps?

CHANGES.rst
amotl: This has been addressed, thanks.

app/app.module.js
amotl: This has been addressed, thanks.

docs/index.rst
seut: > You mean the monitoring and the privileges views? 

No I meant the community edition reference ;) 
`.. _Community Edition: https://crate.io/docs/crate/reference/en/latest/editions.html#cratedb-community-edition`

app/plugins/monitoring/monitoring.css
seut: Ah all other files don't have any header, sorry wasn't aware of this. So ok lets just remove it for now as otherwise we should add a license header in general for everything dedicated. Not sure if that is needed..

docs/index.rst
amotl: Removed. Thanks!

Issue #700, opened 2020-11-24
UI jumps to Tables view when using console 


faymarie:
Dear Craties, 

our team noticed a UI glitch when using the UI console. When working on a query in the console view, the interface frequently jumps back to displaying the first table of the first schema in the "Tables view". We observed this behaviour when working on queries related to subordinate schemas. 

This seems to be unrelated to the query itself. It occurs before and also while editing the query in the UI console, but not after the query was executed. This is especially inconvenient when longer queries get lost.

While editing a query in the console:
![Screenshot 2020-11-24 at 14 00 22](https://user-images.githubusercontent.com/32742984/100099112-43d90900-2e5f-11eb-8f67-7128d523d02e.png)

It jumps to this table view, like here:
![Screenshot 2020-11-24 at 14 04 16](https://user-images.githubusercontent.com/32742984/100098348-38391280-2e5e-11eb-8d64-d5230a8c09a3.png)

**CrateDB Version: 4.2.6**

amotl:
Hi Marie and Naomi,

I tried to track this down and found #304 and #311. There is even a video which tries to explain the rationale for this fix at https://recordit.co/hU5WEMAANc. However, I can't make any sense of it.

I believe the code isn't needed anymore as it also works well without, at least from my perspective.

With kind regards,
Andreas.

amotl:
I think the problem got fixed, so I am closing this issue now. Feel free to reopen if you see fit.

PR #708, opened 2021-02-18
Remove misguided redirects which may lose queries currently being typed

amotl:
Hi there,

@faymarie reported #700 the other day, thank you. This patch might fix it.

I believe the selection of the "current" table/view has gotten new code already. Having redirection code in a function which gets called when registered to the `STATE_REFRESHED` event is probably a bad idea as exactly this might produce the observed behavior.

While everything should usually be fine, I can imagine an edge case where this event handler will not get detached when the page/view object is destroyed by navigating to another page. In this case, the event might fire in the background when a new table is added by another arbitrary operation potentially run by another user. And - there you go - the UI will swap out the current page and display the "Tables" page.

As I haven't been able to observe any noticeable difference in UI behavior after removing those lines of code, I am assuming it will be safe to delete it. However, I will be happy to learn otherwise. Still, I am inclined to believe that if we really need that kind of mechanism, it should be implemented differently instead of issuing a client-side redirect by updating the `location.url` attribute.

With kind regards,
Andreas.

cc @faymarie, @quodt

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,7 +6,9 @@ Changes for Crate Admin Interface
 Unreleased
 ==========
 
-- Align with license change. Enterprise features are now available to everyone.
+- Removed misguided redirects which may lose queries currently being typed.
+
+- Aligned with license change. Enterprise features are now available to everyone.
 
 - Fixed translations. French now works again.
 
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -6,6 +6,8 @@ Changes for Crate Admin Interface
 Unreleased
 ==========
 
+- Remove misguided redirects which may lose queries currently being typed.
```
seut, line 9:
```suggestion
- Removed misguided redirects which may lose queries currently being typed.
```

seut (approved):
Lgtm otherwise

PR #715, opened 2021-03-07
adapt table render function

### LINKED ISSUE #714: Improve render-Function invoked by "Query Table"-Button to only quote identifiers when necessary
Currently only identifiers get surrounded by double quote when the "Query Table"-buttons is pressed, regardless if quoting is really needed. Since this makes it harder to select and delete single identifiers (e.g. double-clicking on identifier), I suggest to follow a similar functionality to quote_ident() also used within crate/postgres.
Further I would also propose to improve the readability of the generated queries by moving `FROM` and `LIMIT` statements to a new line:

Sample Table
```sql
CREATE table quotes (
  ident TEXT,
  "Quotes" TEXT,
  "table" TEXT,
  "TABLE" TEXT,
  "ident quotes" TEXT
);
```

---

Current Output:
```sql
SELECT "ident","Quotes","table","TABLE","ident quotes" FROM "doc"."quotes" LIMIT 100;
```

---

Suggested Output:
```sql
SELECT ident, "Quotes", "table", "TABLE", "ident quotes"
FROM doc.quotes
LIMIT 100;
```

proddata:
add quoteIdent() and isReservedKeyword()

closes #714 

- ~~Add quoteIdent function to table controller~~
- ~~Add isReservedKeyword function to table controller~~
- ~~Adapt render function to use quoteIdent to only quote identifiers when needed~~
- Add new lines before `FROM` and `LIMIT` in render function
- Add whitespace between identifiers
- Use server-side `quote_ident()` function when fetching column-names

## Checklist

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -39,8 +39,6 @@ jobs:
           #       still do it because it will yield tremendous speed improvements.
           #       By adding the `node-version` to the cache key, it should work.
           key: ${{ runner.os }}-node-${{ matrix.node-version }}-${{ hashFiles('**/package-lock.json') }}
-          restore-keys: |
-            ${{ runner.os }}-node-
 
       - name: Install package
         run: npm install
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -21,6 +21,10 @@ Unreleased
 
 - Fixed missing whitespace.
 
+- When navigating from table view to console, apply server-side ``quote_ident``
+  function to get better usability for column names within the manifested SQL
+  statement. Also, added appropriate newlines.
+
 
 2020-10-12 1.17.0
 =================
--- a/app/scripts/controllers/tables.js
+++ b/app/scripts/controllers/tables.js
@@ -112,7 +112,7 @@ angular.module('tables', ['stats', 'sql', 'common', 'tableinfo', 'events'])
 
         const COLUMN_QUERY = `
 SELECT
-    column_name,
+    quote_ident(column_name) as column_name,
     upper(data_type) AS data_type,
     is_generated = 'ALWAYS' as is_generated,
     generation_expression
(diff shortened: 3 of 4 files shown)


amotl:
Hi Georg,

thank you very much for this patch. I wonder if the list of keywords already implemented at [1] might be able to be reused in any way?

With kind regards,
Andreas.

[1] https://github.com/crate/crate-admin/blob/b84f8095576c04d3da08ff25f7737fded4651c6d/app/scripts/controllers/console.js#L392-L500

amotl:
> Happy do adjust, but probably need some hints, since my angular knowledge is lacking :)
> From my understanding one would need to move that part to a component, as directives should be self-contained.

Just wanted to make sure the outcome would fit your intentions. I can help on the technical side as I am currently working on the code base anyway.

proddata:
I made some small changes to use the server-side quote-ident function when fetching the column names in the first place and remove the client-side quote-ident.

This won't (yet) fix the quoting of namespace / table name, but would be a good step forward

This leads to quotes also showing up on the table-page e.g.
image

but I'd consider that fine

amotl:
@proddata proposed another change to just use the `quote_ident()` of CrateDB itself for this job. In that matter, the keywords would not even have to be fetched and operated on. What do you think @mfussenegger?

proddata:
I made some small changes to use the server-side quote-ident function when fetching the column names in the first place and remove the client-side quote-ident.

This won't (yet) fix the quoting of namespace / table name, but would be a good step forward

This leads to quotes also showing up on the table-page e.g.
image

but I'd consider that fine

amotl (approved):

amotl:
Hi Georg,

I polished your patch a bit and made the tests on CI work. I believe it will be good to go now.

With kind regards,
Andreas.

PR #724, opened 2021-06-11
Add missing data type keywords

norosa:
While editing [data-types.rst][1] in the main CrateDB Reference, I noticed that the following data types do not get automatically uppercased by the console:

- `int2`
- `int4`
- `int8`
- `name`
- `oid`
- `oidvector`
- `regclass`
- `regproc`
- `timestamptz`
- `timetz`

This commit adds these data types to the list of keywords.

[1]: https://crate.io/docs/crate/reference/en/4.5/general/ddl/data-types.html


amotl (approved):
Thank you @norosa!

@mfussenegger and @seut: There is no remote command to inquire the list of keywords from CrateDB itself yet, right?

mfussenegger:
> There is no remote command to inquire the list of keywords from CrateDB itself yet, right?

```
select * from pg_get_keywords()
```


Not all data types are automatically keywords. In fact, most aren't.

Looks like the list of keywords in the admin-ui already contains a lot of things that are not keywords in CrateDB.

amotl:
@mfussenegger: Thank you. We can discuss how this situation can be improved the next time we meet? Do you have any specific suggestions or is there already some plan from the past?

mfussenegger:
In crash there was a change to load the keywords dynamically if the server side information is available: https://github.com/crate/crash/commit/a01a352027865b9142931f0aceb90d76bfe023e3

In the admin interface we could do the same (and without version/availability checks, because the admin interface is version bound to specific CrateDB versions anyways)

norosa:
@mfussenegger I noticed the same thing. as a rule, I capitalize _reserved keywords_ per the SQL standard. however, I wasn't aware of the `pg_get_keywords` function. good to know

while working on https://github.com/crate/crate/pull/11418, I was trying to determine what to capitalize. I had the Admin UI open and was performing experiments already, so I was quick to arrive at the conclusion that the examples in the docs should match the capitalization in the Admin UI console (_even if_ they weren't reserved keywords, per my usual rule of thumb)

I was planning to do a follow-up PR for crash (thinking it was likely to work in a similar way to the Admin UI) but I see that that isn't possible

should we merge this PR _as is_ and create a new issue for implementing `pg_get_keywords` functionality, or should we close it?

amotl:
If this patch should go in in general, not taking other actions into account, I would like to include it into the upcoming release.

amotl:
Hi again,

Mergify apparently doesn't do its job here. So, after integrating this, #729 will probably have to be adjusted. Do you agree, @norosa and @proddata?

With kind regards,
Andreas.

amotl:
Is there a final verdict on whether this is appropriate to add or not?

/cc @proddata, @mfussenegger, @seut

mfussenegger:
I think https://github.com/crate/crate-admin/issues/737 is the way to go. As was already pointed out - most data types are actually not keywords.

norosa:
ack

PR #725, opened 2021-06-14
Remove analytics (UDC, Segment) and other kinds of home calling

amotl:
Hi there,

CrateDB already includes a [Usage Data Collector (UDC)](https://crate.io/docs/crate/reference/en/latest/admin/udc.html) component. There have been some additional components in the Admin UI which were thought to deliver those kinds of features:

- Support chat integration with Intercom
- Intercom integration through Segment, also providing tracking analytics
- Compare running/recent versions with https://crate.io/versions.json
- Populate the wrench menu with content from https://crate.io/feed/menu.json

As we are a) phasing out Intercom and b) the other features have never actually been used actively, I believe it is for the better to remove those features altogether. In that manner, the user navigating in Admin UI will never see the browser making calls to any 3rd party locations whatsoever.

With kind regards,
Andreas.


michaelkremmel (approved):
Matthias had a look at the code. For me the changes are okay too.

msbt:
Removing Segment will result in the loss of Google Analytics data (which according to @michaelkremmel hasn't been working for quite a while and a webhook pointing to https://udc.crate.io/uid with no credentials/tokens/secret, so I reckon this doesn't do anything either. If you can confirm that @amotl, I'll gladly approve

msbt (approved):
👍

norosa:
not sure about removing these bits:

* Compare running/recent versions with https://crate.io/versions.json
* Populate the wrench menu with content from https://crate.io/feed/menu.json

the first one is in use, in the sense that it notifies users when a newer version of CrateDB is available, and as far as I know, this feature still works. https://crate.io/versions.json is actively updated. I don't understand why we would remove this functionality. it seems useful to me. @mfussenegger probably has input here

re https://crate.io/feed/menu.json: this seems like a useful thing to keep around. why remove it?

amotl:
Hi Naomi,

the why was: Remove everything which calls home in order to better respect the user's privacy and to remove any obstacles when running CrateDB in constrained environments.

- The dynamic menu (feed/menu.json) might sound useful, but apparently it does not actually get used. I wonder whether it was in the past?
- Let me know if you believe the version update note feature should stay. Personally, I never experienced it in action but maybe I overlooked it or my browser blocked it?

With kind regards,
Andreas.

seut (approved):
Lgtm. 👍 for removing the analytics calls. I've no strong opinion on the update notifications, but if they do not work, we should remove or fix. For now I'd prefer first one, we can re-add this anyhow if wanted.

PR #726, opened 2021-06-14
Update several dependency packages

amotl:
Hi there,

this patch upgrades some dependency packages to more recent versions in order to modernize the environment and mitigate some security notices revealed by `npm audit`.

With kind regards,
Andreas.


CHANGES.rst
```diff
@@ -6,6 +6,8 @@ Changes for Crate Admin Interface
 Unreleased
 ==========
 
+- Update several dependency packages
```
seut, line 9: As these are afaik no user facing changes, I wouldn't add any changes entry.

seut (approved):
👍 Just added a comment related to the changes entry.

PR #728, opened 2021-06-16
Rename default branch to "main"

amotl:
Hi there,

following some other repositories, this patch will rename the default branch to `main`. Is there any other thing to do beyond merging this and actually renaming the branch on GitHub, @norosa?

With kind regards,
Andreas.


norosa (approved):
@amotl I think you may need to update the RTD config, but other than that, I can't think of anything

PR #740, opened 2021-08-31
Fix generated queries for views containing identifiers that need to be quoted

proddata:

Fix generated queries for views containing identifiers that need to be quoted.
Identifiers that need to be quoted (e.g. with upper case letters or reserved keywords) were double quoted twice

This bug was introduced with https://github.com/crate/crate-admin/pull/715 😬 

@jayeff 

## Checklist


amotl (approved):
Thank you!

PR #742, opened 2021-09-05
side-nav: fix monitor icon color + make slightly smaller

proddata:
side-nav: fix monitor icon color + make slightly smaller + hover-effect

## Checklist

### DIFF
--- a/app/styles/views/_nav.scss
+++ b/app/styles/views/_nav.scss
@@ -17,15 +17,29 @@
 	cursor: pointer;
 
 	width: 100%;
-	padding-top: 20px;
-	padding-bottom: 20px;
+	padding-top: 18px;
+	padding-bottom: 18px;
 	text-align: center;
+}
+
+.cr-side-nav__item:hover {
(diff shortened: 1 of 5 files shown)


app/styles/views/_nav.scss
```diff
@@ -17,15 +17,29 @@
...
 .cr-side-nav__item > img {
 	opacity: 0.4;
 
 }
+
+.cr-side-nav__item:hover > img{
+	opacity: 1.0;
+}
+
+.cr-side-nav__item--active  {
```
msbt, line 38: can we make that 3px solid ... so the formal syntax is in order (i.e. line-width line-style color)?
https://developer.mozilla.org/en-US/docs/Web/CSS/border#formal_syntax for reference

other than that, looks good!

msbt (commented):
see below, not sure if this is really necessary, but I have never seen that in that order ;)

msbt (approved):
lgtm!

amotl (approved):
Thank you!

PR #743, opened 2021-09-05
Move query monitoring to overview and remove dedicate monitoring section

### LINKED ISSUE #733: Move query monitoring to main overview
Since 4.5.0 all enterprise features (including monitoring) are included in all versions of CrateDB.
Therefore I would propose to combine the former **Monitoring view** with the main **overview**

e.g.
**Alternative A**
image

**Alternative B**
image

proddata:

- Move query monitoring to overview and remove dedicate monitoring section.
- Aggregate QPS and Query duration by 10 second intervals to smoothen graph and reduce total number of points to be drawn (The former high number lead to performance problems when resizing the browser window.
- Dropped logic of buffering old results, as it not needed with the the 10 sec aggregation anymore and it lead to duplicated entries
- Slightly adjust the graph buttons to align to the right and break behaviour

closes #733

Screenshot 2021-09-19 at 16 05 31

Happy for feedback @amotl @seut @mfussenegger 

## Checklist


mfussenegger:
Seems good to me. Maybe  @belringer also has some inputs?

proddata:
Actually I am unsure, if the "Query Per Second" and "Query Speed" isn't a bit misleading overall to show to non-superusers (`crate`), as it only includes the own queries for non-superusers.

technically this could definitely could use some clean up and deduplication (i.e. load / monitoring graphs are implemented separately)

amotl (approved):
Thank you for streamlining the user interface after the differentiation of the former enterprise features have been removed. As @mfussenegger also generally agrees, I believe it will be good to go.

belringer:
Looks good to me. Since the monitoring isn't too complex and it is placed on the front page, I don't see the need for a separate section. Plus the front page is already showing monitoring related information.

proddata:
@belringer Do you think it makes sense to make them a bit smaller and maybe display 2 or 3 graphs next to each other.

--

Plan would be to clean it up a bit more on the implementation side :)

belringer:
@proddata Sure, maybe something like this:
image
But then I would also suggest a more refined implementation of the chart toggles, as they break a bit strangely (see screenshot above).

proddata:
@belringer 
how about so 😬  ... would be a bit more in line with the Cloud UI

image

proddata:
I would say it is ready for review now :)

belringer (approved):
Looks good to me.

amotl:
Thank you. Let's get it merged in order to cut another release?

CHANGES.rst
```diff
@@ -8,7 +8,13 @@ Unreleased
 - Side-Navigation: Adjusted size sligthly, added hover-effect on selected item,
   added border on active item and fixed monitoring icon
 
+<<<<<<< HEAD
 - Added ability to correctly display nested arrays in OBJECT(IGNORED) columns
+=======
+- Move Monitoring to Overview and remove dedicated section. Monitoring now 
+  always show last 15 minutes of QPS and query duration in 10 sec intervals.
+  Slighty adjusted appearance of graphs.
+>>>>>>> 504871e (Move query monitoring to overview and remove)
```
amotl, line 17: It looks like conflict markers slipped in here.

app/static/i18n/de.json
```diff
@@ -43,7 +43,9 @@
         "AFFECTED_NODES": "Betroffene Knoten",
         "INFO": "INFO",
         "WARNING": "WARNUNG",
-        "CRITICAL": "KRITISCH"
+        "CRITICAL": "KRITISCH",
+        "QUERIES_PER_SECOND": "Anfragen Pro Sekunde",
+        "QUERY_SPEED": "Abfragegeschwindigkeit (ms)"
```
amotl, line 48: Would this better be called "QUERY_DURATION"? Not totally sure, I would also be reasonably fine with SPEED.
```suggestion
        "QUERY_DURATION": "Abfragedauer (ms)"
```

CHANGES.rst
proddata: Thanks ... VS Code played tricks on me 😬

app/static/i18n/de.json
proddata: I am indifferent, but would need good translations for EN,DE,FR,ES :D 
I only moved it from the monitoring plugin.

amotl: Then let's keep it like it was.

PR #757, opened 2022-01-13
Use NPM 8.3 and provide manual overrides for some packages with updates

amotl:
Hi,

there have been some Dependabot alerts on this repository. Most if not all of them are related to packages transitively pulled in from development dependencies.

image

In order to improve the situation, the patch uses the [`overrides` mechanism](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides) introduced by NPM 8.3.

With the upgrade to NPM 8, this also upgrades to lock file version 2.

With kind regards,
Andreas.


seut (approved):
👍

PR #778, opened 2022-09-11
Fit console query results in viewport

proddata:
- **Fit console query results in viewport**
  This should make it easier to work with wider result-sets, as one does not longer need to scroll to the bottom of the page to use the horizontal scrollbar
- **Fix: Adjust scrollbar corner color to fit with theme**
  was white before / didn't fit with rest of theme

image

## Checklist

 - [ x ] [CLA](https://crate.io/community/contribute/cla/) is signed

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,9 @@ Changes for Crate Admin Interface
 Unreleased
 ==========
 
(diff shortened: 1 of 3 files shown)


seut (approved):
👍 Thank you.

PR #796, opened 2022-12-02
Chore: Update to Node.js 16

amotl:
Following up on #794 and #795, this patch also updates the sandbox/build/CI environment from Node.js 14 to Node.js 16.

-- https://endoflife.date/nodejs


hammerhead (approved):
Great to see this upgraded!

PR #797, opened 2022-12-02
Sandbox/CI: Update to Node.js 18

amotl:
### About

Following up on #796, this patch updates the sandbox/build/CI environment to use Node.js 18.

### Caveats

`NODE_OPTIONS=--openssl-legacy-provider` was needed to satisfy `webpack@4`.

- [Error message "error:0308010C:digital envelope routines::unsupported"](https://stackoverflow.com/a/69699772)
- https://github.com/webpack/webpack/issues/14532#issuecomment-947012063
- https://github.com/nodejs/node/issues/40455#issuecomment-1123071147

The real solution is to update to `webpack@5`.

- https://github.com/webpack/webpack/issues/14532#issuecomment-1324340823

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       matrix:
         os: [ ubuntu-latest ] #, macos-latest, windows-latest ]
-        node-version: [ '16' ]
+        node-version: [ '18' ]
 
     name: Node.js ${{ matrix.node-version }} on OS ${{ matrix.os }}
     steps:
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -21,7 +21,6 @@ pull_request_rules:
       queue:
         method: rebase
         name: default
-        rebase_fallback: none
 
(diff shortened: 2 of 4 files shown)


amotl:
Why is the process on Node.js 18 so slow, even when using a warm package cache?

### Node.js 16
> added 1232 packages, and audited 1233 packages in 16s

-- https://github.com/crate/crate-admin/actions/runs/3603406093/jobs/6071565432#step:4:28

### Node.js 18
> added 1232 packages, and audited 1233 packages in 2m

-- https://github.com/crate/crate-admin/actions/runs/3603697467/jobs/6072416418#step:4:28

amotl:
It looks like the slowness was related to the attempt to upgrade to NPM 9.x, with 90eb385be9. The run 3677157095 [^1] shows that without it, _Install packages_ is back to 17s.

[^1]: https://github.com/crate/crate-admin/actions/runs/3677157095/jobs/6218786856

hammerhead (approved):

PR #809, opened 2023-02-15
Fixed monitoring statement for compatibility with CrateDB 5.2

amotl:
CrateDB 5.2 needs `AVG(ended::bigint - started::bigint)` instead of `AVG(ended - started)`.

### DIFF
--- a/app/scripts/controllers/overview.js
+++ b/app/scripts/controllers/overview.js
@@ -440,7 +440,7 @@ angular.module('overview', ['stats', 'checks', 'ngSanitize', 'events', 'filters_
       return 'SELECT CURRENT_TIMESTAMP AS last_timestamp, ' +
         '(ended / 10000) * 10000 + 5000 AS ended_time, ' +
         'COUNT(*) / 10.0 AS qps, ' +
-        'AVG(ended - started) AS duration, ' +
+        'AVG(ended::bigint - started::bigint) AS duration, ' +
         'UPPER(regexp_matches(stmt,\'^\\s*(\\w+).*\')[1]) AS query_type ' +
         'FROM sys.jobs_log ' +
         'WHERE ended > now() - (\'15 minutes\')::interval ' +
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,9 @@ Changes for Crate Admin Interface
 Unreleased
 ==========
 
+- Fixed monitoring statement for compatibility with CrateDB 5.2,
+  ``AVG(ended - started)`` becomes ``AVG(ended::bigint - started::bigint)``.
+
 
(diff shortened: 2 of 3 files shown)


matriv (approved):
Thank you!

Did you verify there is no other subtraction of timestamps used?

app/scripts/controllers/overview.js
```diff
@@ -440,7 +440,7 @@ angular.module('overview', ['stats', 'checks', 'ngSanitize', 'events', 'filters_
       return 'SELECT CURRENT_TIMESTAMP AS last_timestamp, ' +
         '(ended / 10000) * 10000 + 5000 AS ended_time, ' +
         'COUNT(*) / 10.0 AS qps, ' +
-        'AVG(ended - started) AS duration, ' +
+        'AVG(ended::bigint - started::bigint) AS duration, ' +
         'UPPER(regexp_matches(stmt,\'^\\s*(\\w+).*\')[1]) AS query_type ' +
         'FROM sys.jobs_log ' +
         'WHERE ended > now() - (\'15 minutes\')::interval ' +
```
amotl, line 446: > Did you verify there is no other subtraction of timestamps used?

What about this spot?

```sql
WHERE ended > now() - (\'15 minutes\')::interval
```

app/scripts/controllers/overview.js
```diff
@@ -440,7 +440,7 @@ angular.module('overview', ['stats', 'checks', 'ngSanitize', 'events', 'filters_
       return 'SELECT CURRENT_TIMESTAMP AS last_timestamp, ' +
         '(ended / 10000) * 10000 + 5000 AS ended_time, ' +
```
amotl, line 441: What about this spot?
```sql
(ended / 10000) * 10000 + 5000 AS ended_time
```

app/scripts/controllers/overview.js
matriv: this shouldn't be a problem it's `timestamp - interval` which returns `timestamp`

amotl:
What about this?

```sql
SELECT CURRENT_TIMESTAMP AS last_timestamp
```

app/scripts/controllers/overview.js
proddata: That is just a timestamp, all good.

That probably could be changed to use `date_bin` ...
```sql
date_bin('10 seconds'::INTERVAL, ended, 0) + 5000 AS ended_time
```

... but not really necessary

amotl: Thanks for the suggestion. Feel free to submit another patch, or let me know if I should do it. If you think it's not required _now_, let's keep this very patch minimal.

Issue #819, opened 2023-10-11
Links pointing to Crate.io 


michaelkremmel:
Hi,
I've realized that some links in the admin UI point to crate.io. Since the domain was moved to cratedb.com, it would be great to change those links at some point (now all of them are redirected correctly). I've also added some proposals to change some of the corresponding copy:

In the section **/help**
![grafik](https://github.com/crate/crate-admin/assets/74347398/6a32ac22-ed99-4de6-a1c1-bb61cea45345)

<table>
<tr><td>Existing</td><td>New</td><tr>
<tr><td>Ask Crate.io engineers in our Discourse community.</td><td>Join our community to get answers to your technical questions and discuss other topics.</td><tr>
<tr><td>https://community.crate.io</td><td>https://community.cratedb.com</td><tr>
<tr><td></td><td></td><tr>
<tr><td>https://crate.io/docs/</td><td>https://cratedb.com/docs/crate/reference/en/latest/</td><tr>
<tr><td></td><td></td><tr>
<tr><td>Enterprise support</td><td>Support plans</td><tr>
<tr><td>For enterprise support, please check our pricing page.</td><td>Get assistance from our engineering team. Check our support plans.</td><tr>
<tr><td>https://crate.io/enterprise/</td><td>https://crate.io/product/pricing/support-plans </td><tr>
</table>

When clicking the **cogwheel**
![grafik](https://github.com/crate/crate-admin/assets/74347398/c96aaae0-f2f3-4f89-ac3b-e2e189cedd8d)

<table>
<tr><td>Existing</td><td>New</td><tr>
<tr><td>https://crate.io/docs/en/5.4/ </td><td>https://cratedb.com/docs/crate/reference/en/5.4</td><tr>
<tr><td></td><td></td><tr>
<tr><td>Schedule a call with a Crate.io engineer</td><td>Schedule a call with a CrateDB engineer</td><tr>
<tr><td>https://crate.io/demo?utm_source=adminui&utm_medium=browser&utm_term=&utm_content=demolink&utm_campaign=newsfeed&ajs_event=clicked_demo_link </td><td>https://cratedb.com/contact </td><tr>
<tr><td></td><td></td><tr>
<tr><td>https://crate.io/blog/</td><td>https://cratedb.com/blog </td><tr>
</table>

amotl:
Hi Michael,

GH-822 and GH-823 bring in the requested adjustments, and a bit more copy-editing.

[...]

PR #823, opened 2023-10-20
Rename "Crate" to "CrateDB", and adjust a few more texts

amotl:
## About

A few more text edits suggested at GH-819, and more.


matriv (approved):
LGTM, but why did you remove those entries from the `LICENCE`?
Maybe wait for another approval for this ^^ ?

amotl:
> Why did you remove those entries from the LICENSE file?

I've trimmed a few of those in other repositories already. In this case, I just moved it to GH-824, in order to discuss later.

PR #824, opened 2023-10-22
Chore: Trim LICENSE file

amotl:
## About

I think there are lots of unneccessary details within the LICENSE file. If you think we should keep some of them, we can re-add them to the NOTICE file, like other projects are doing it.


seut (approved):
👍

Issue #861, opened 2024-10-25
Warnings after recent webpack 5 and loader updates


amotl:
## Status
Software tests succeed, yet log output includes warnings.
```
WARNING in ./app/styles/styles.scss 22:17-24
export 'default' (imported as 'content') was not found in '!!../../node_modules/mini-css-extract-plugin/dist/loader.js!../../node_modules/css-loader/dist/cjs.js??ruleSet[1].rules[1].use[2]!../../node_modules/sass-loader/dist/cjs.js??ruleSet[1].rules[1].use[3]!./styles.scss' (module has no exports)
 @ ./app/app.module.js 4:0-30
```
-- https://github.com/crate/crate-admin/actions/runs/11517066743/job/32061002353?pr=860#step:5:11

Apparently, the warnings are originating at the webpack bundler. They appeared with 360b5ebe55706aad0471e28c1c7aaba4571b9a0b and e62b9cab04bdfa0ff471d1e41e49da23a0053d9f.

amotl:
### Possible References
- https://github.com/webpack-contrib/mini-css-extract-plugin/issues/613
- https://github.com/webpack/webpack/issues/12558

PR #862, opened 2024-10-25
Sandbox: Fix update to webpack 5

amotl:
## Problem
After updating to webpack 5, the build/test invocation introduced a few warnings.
- GH-861

## Solution
Follow the advice to not use `style-loader` and `mini-css-extract-loader` together.

- https://github.com/webpack-contrib/mini-css-extract-plugin/issues/613#issuecomment-707653736


surister:
Does make sense that we maintain this proj in any capacity?

PR #865, opened 2024-10-28
CI: Add Dependabot configuration file `dependabot.yml`

amotl:
This makes it possible to let it check for updates manually, and to refresh the list of currently active security warnings.

-- https://stackoverflow.com/a/77502174


amotl:
After merging, the job is _running now_.

image

-- https://github.com/crate/crate-admin/network/updates/10095311/jobs

PR #871, opened 2024-10-28
CI: Use both Node.js 20 and 22

amotl:
Mostly to check which builds faster on CI/GHA.


amotl:
> Mostly to check which builds faster on CI/GHA.

It takes the same amount of time. Closing.

PR #891, opened 2025-05-13
Docs: remove in page toc

bmunkholm:
The in-page TOC is redundant as it's also present to the right side.
"Screenshot" headline seems a bit silly, so removed that.


amotl (approved):
> The in-page TOC is redundant as it's also present to the right side.

Yeah, the secondary navigation element (rightnav) is a new element compared to how the documentation layout and structure had been conceived originally, many years ago. In this spirit, I think we can remove many on-page TOC elements across the board. Thank you so much!

PR #894, opened 2025-05-19
Remove documentation for Help page

bmunkholm:

Removed documentation of the Help page. Half the text was wrong (you can import tweets). The remaining text just say that it contains "help resources".

There simply is no value add for users by having this view "explained" in the documentation. It just seems to be there for the sake of completenes.


seut (changes requested):
I actually don't understand this PR at all.
It only removes the documentation about the help sceen.
Additionally, the help screen does not mention any obsolete "import tweet" content but just links to where to get help and all is still valid.

bmunkholm:
@seut The help screen documentation that I suggest to remove says:
> comes with a *help screen* that allows you to import some tweets for testing purposes.

That seems to be incorrect. I don't see anywhere that the Help screen has this feature?

If we correct that text, what's left besides the screenshoot is just:
> The CrateDB Admin UI comes with a help screen that links to important CrateDB support resources.

Which seems ... rather obvious?

As a user I don't get any value out of that page. And then its just fluff that I think we should trim away.

seut:
> @seut Right, I can see how this can be misunderstood. I've updated title and description. This is just about documentation. I only suggest to remove the help screen documentation, no changes to the application.

Ah ok, thanks for the clarification. Right, I agree, this documentation does not add much value, lets remove.

> The help screen documentation that I suggest to remove says:
> 
> > comes with a _help screen_ that allows you to import some tweets for testing purposes.
> 
> That seems to be incorrect. I don't see anywhere that the Help screen has this feature? Am I missing something?

Yes, this feature was removed long time ago.

seut (approved):

